PR #5, opened 2025-08-22
docs: add the talos docs

Iheanacho-ai:


docs.json
```diff
@@ -43,6 +43,301 @@
   },
   "navigation": {
     "tabs": [
+      {
```
rothgar, line 46: Can you move these to a yaml file and generate the docs.json instead? If there's a feature missing from the generator we can work on that. All of this looks pretty standard.

talos/v1.10/build-and-extend-talos/cluster-operations-and-maintenance/watchdog.mdx
```diff
@@ -0,0 +1,69 @@
...
+Talos Linux now supports hardware watchdog timers configuration.
+Hardware watchdog timers allow to reset (reboot) the system if the software stack becomes unresponsive.
+Please consult your hardware/VM documentation for the availability of the hardware watchdog timers.
+
+## Configuration
+
+To discover the available watchdog devices, run:
+
+```shell
+$ talosctl ls /sys/class/watchdog/
```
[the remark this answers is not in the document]
Iheanacho-ai: No I wasn't. I feel like that would take way too much time. I think it would be better to remove them as time goes by.

docs.json
Iheanacho-ai: done

omni.yaml
```diff
@@ -1,5 +1,239 @@
 navigation:
   tabs:
+    - tab: "Talos"
```
rothgar, line 3: Since you added the other talos yaml file you need to remove these from the omni.yaml.

Iheanacho-ai: Done, I did remove it earlier but I think when I rebase, it erased my changes

PR #40, opened 2025-09-27
fix: OIDC provider authentication example

joshdev-de:
While following the documentation I discovered that the --auth-oidc-provider-url flag is missing in the example while being required to run omni. Some missing backslashes are added as well.


rothgar (approved):
Thanks!

PR #45, opened 2025-10-01
docs: oidc auth through tsidp

bo0tzz:


docs.json
bo0tzz, line 1: I've run `make docs.json` but I think it did something weird here.

PR #50, opened 2025-10-06
docs: add reference docs converted from talosctl

rothgar:
Update docs-convert to handle more edge cases and formatting and add a comment to note docs are automatically generated. Remove examples from inside reference tables. they don't render properly on mintlify and currently only work for hugo.


talos-v1.11.yaml
```diff
@@ -201,14 +201,31 @@ navigation:
         - group: "Reference"
           folder: "talos/v1.11/reference"
           pages:
+            - "v1alpha1/config.mdx"
```
rothgar, line 204: I wanted to bring this up to the top level of the references because it's the most common and important configuration to reference. All other configuration folders were flattened into 1 page if there was only 1 configuration in that group.

image

We could keep the groups but then someone has to expand the section before they can see the config documents in that group (IMO it's annoying)

rothgar:
Preview link https://siderolabs-fe86397c-1-11-reference.mintlify.app/talos/v1.11/overview/what-is-talos

PR #71, opened 2025-10-14
docs: added job that uses configmap with helm values to install cilium

basbremer:
The current documentation provides an option to create a Kubernetes Job that handles the installation of Cilium.
Unfortunately, you need to provide all configuration options directly within this Job as arguments.

**What changed?**
This PR enhances the documentation by introducing a Job that allows you to provide the Cilium Helm values separately.
The Helm values are supplied via a ConfigMap and mounted as a file in the Job.

**Why it's needed?**
In my opinion, this gives better visibility of the desired configuration options for Cilium, especially when you have many of them.

**Screenshot**
The part shown in the image below is added to the existing Cilium docs:
Screenshot 2025-10-28 at 16 16 21

### DIFF
--- a/public/kubernetes-guides/cni/deploying-cilium.mdx
+++ b/public/kubernetes-guides/cni/deploying-cilium.mdx
@@ -535,6 +535,6 @@ For more details see [the discussion here](https://github.com/siderolabs/talos/p
   This is expected, you can workaround it by adding the `pod-security.kubernetes.io/enforce=privileged` [label on the namespace level](../security/pod-security).
 
 - Talos has full kernel module support for eBPF, See:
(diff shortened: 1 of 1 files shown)


Iheanacho-ai:
@smira what do you think of the code quality of this PR, do you think its good to merge?

smira (commented):
I don't like any of the ways except for the official Cilium CLI way. I think we should stop even publishing this and link to Cilium docs. 

These Kubernetes jobs are all hacks which I don't want to support in any way. Of course people are free to use whatever way they feel works great, but it shouldn't be the way we recommend.

Iheanacho-ai:
Alright. I will close this PR

PR #79, opened 2025-10-20
docs: list sensitive talos resources when omni-managed

alongwill:
Show the list of sensitive Talos resources that cannot be viewed for Omni managed clusters.

### DIFF
--- a/public/omni/getting-started/how-to-install-talosctl.mdx
+++ b/public/omni/getting-started/how-to-install-talosctl.mdx
@@ -54,3 +54,11 @@ However, users with at least the [Operator](https://omni.siderolabs.com/referenc
 | `talosctl restart`                             |
 | `talosctl service <id> [stop\|start\|restart]` |
 | `talosctl shutdown`                            |
+
+## Sensitive resources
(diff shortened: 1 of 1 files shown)


public/omni/getting-started/how-to-install-talosctl.mdx
```diff
@@ -54,3 +54,11 @@ However, users with at least the [Operator](https://omni.siderolabs.com/referenc
 | `talosctl restart`                             |
 | `talosctl service <id> [stop\|start\|restart]` |
 | `talosctl shutdown`                            |
+
+## Sensitive resources
+
+Omni prevents viewing machine configuration which might contain sensitive PKI. To view the list of resources that cannot be viewed you can use this command:
```
Iheanacho-ai: i added this comment yesterday but for some reason it has been pending since, but here are my thoughts:

i personally think that there are too many "view" in this paragraph

would this be a better rewrite?

'Omni restricts access to machine configurations that may contain sensitive PKI information. To see which resources are restricted, run the following command:'

PR #139, opened 2025-11-04
docs: review the install omnictl

Iheanacho-ai:


public/omni/getting-started/install-and-configure-omnictl.mdx
```diff
@@ -54,24 +88,41 @@ If you would like to merge the `omniconfig.yaml` with an existing configuration,
...
 
 ```bash
 omnictl get clusters
 ```
 
-If the browser window does not open automatically, it can be opened manually by copying and pasting the URL into a web browser:
+A browser window should open automatically for sign-in. If it doesn’t, copy and paste the URL manually:
 
 ```bash
 BROWSER=echo omnictl get clusters
```
rothgar: @Unix4ever isn't there a line like `--skip-open-browser` they should add to their omni config to make it not open the browser automatically?

PR #148, opened 2025-11-07
docs: add talos config overrides

### LINKED ISSUE #81: Document talos config that doesn't work with Omni
There are some features of the Talos config that don't work with or are unnecessary with Omni. We need to document what features those are and why they don't work.

Examples include:

certificates (talos, etcd, or kubernetes) since Omni manages certificates for you
disk encryption - I'm not sure if this is still possible to set
cluster VIP - I don't think this works since Omni is the external endpoint

Iheanacho-ai:
fixes #81

### DIFF
--- a/omni.yaml
+++ b/omni.yaml
@@ -65,6 +65,7 @@ navigation:
             - "restore-etcd-of-a-cluster-managed-by-cluster-templates.mdx"
             - "using-audit-log.mdx"
             - "importing-talos-clusters.mdx"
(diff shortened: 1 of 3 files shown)


rothgar:
You should link cluster imports here and mention these fields will be changed when the cluster is imported into omni

https://docs.siderolabs.com/omni/cluster-management/importing-talos-clusters

rothgar (approved):

PR #159, opened 2025-11-12
docs: add docs update on v1.12 image cache

Iheanacho-ai:


public/talos/v1.12/configure-your-talos-cluster/images-container-runtime/image-cache.mdx
```diff
@@ -150,6 +150,46 @@ Only images baked into the ISO or USB are copied. There is no way to push arbitr
...
+```bash
+talosctl image cache-serve \
+  --dir ./_out/image-cache.oci \
+  --listen 0.0.0.0:5000
+```
+
+Once running, the cache can be accessed via:
+
+```url
+http://<host>:5000/
```
shanduur: It's worth to add how it is actually intended to be used in Talos, e.g. using the patch:


1. Generating minimal image cache can be done using following chain of commands:
    ```bash
    talosctl images default |\
      talosctl images cache-create --images=- --image-cache-path=/tmp/cache --layout=flat
    ```

1. Additionally, it needs a set of certificates. You can generate them (and required patch - see `image-cache-patch.yaml`) using the following command:
    ```bash
    talosctl image cache-cert-gen --advertise-address=172.20.0.1 \
      --tls-ca-file=/tmp/ca.crt \
      --tls-cert-file=/tmp/tls.crt \
      --tls-key-file=/tmp/tls.key
    ```

1. We now can start the image cache:
    ```bash
    talosctl image cache-serve \
      --image-cache-path=/tmp/cache \
      --address=172.20.0.1:12000 \
      --tls-cert-file=/tmp/tls.crt \
      --tls-key-file=/tmp/tls.key
    ```

1. `image-cache-patch.yaml` - Talos requires HTTPS to pull installer image. In airgapped setups, images are hosted in an internal OCI registry using a self-signed or private TLS certificate. Because Talos does not trust this certificate by default, it will fail with `x509: certificate signed by unknown authority`. This patch adds the certificate of private registry to Talos trusted CAs so it can securely pull images.
    ```yaml
    apiVersion: v1alpha1
    kind: TrustedRootsConfig
    name: image-cache-ca
    certificates: |
      -----BEGIN CERTIFICATE-----
      ...
      -----END CERTIFICATE-----
    ```

1. Talos and Kubernetes components normally pull images from public registries like `docker.io`, `ghcr.io`, and `registry.k8s.io`. These are unreachable in airgapped environments. This patch tells Talos to redirect all image pulls to local registry mirror at `https://172.20.0.1:12000`. This is the same registry as the one hosting installer image, thus we need to specify HTTPS and not HTTP as it's often showed in Talos docs.
    ```yaml
    machine:
      registries:
        mirrors:
          docker.io:
            endpoints:
              - https://172.20.0.1:12000 # note that it is HTTPS not HTTP
          gcr.io:
            endpoints:
              - https://172.20.0.1:12000 # note that it is HTTPS not HTTP
          ghcr.io:
            endpoints:
              - https://172.20.0.1:12000 # note that it is HTTPS not HTTP
          registry.k8s.io:
            endpoints:
              - https://172.20.0.1:12000 # note that it is HTTPS not HTTP
    ```

shanduur (changes requested):

public/talos/v1.12/configure-your-talos-cluster/images-container-runtime/image-cache-registry-mirror.mdx
```diff
@@ -0,0 +1,103 @@
...
+---
+
+Talos can bundle container images into an image cache and use it locally on each node.
+In air-gapped or restricted environments, you can also serve this cache over HTTPS and configure Talos to use it as a registry mirror.
+
+To serve the image cache over HTTPS:
+
+## Step 1. Create the image cache
+
+First, build a list of images and create the cache. This example builds a minimal Talos image cache. To learn how to create an image cache, see the [Image cache documentation](./image-cache)
```
shanduur: Do we need to add documentation about `flat` image cache there?

PR #210, opened 2025-12-08
docs: add different infrastructure provider

### LINKED ISSUE #172: Add missing infrastructure providers
We should add the new infrastructure providers to the docs. We don't need full guides to show how to set them up (some of them are in alpha state), but users should be able to find the relevant github repos from the docs.

Iheanacho-ai:
fixes #172

### DIFF
--- a/public/omni/infrastructure-and-extensions/infrastructure-providers.mdx
+++ b/public/omni/infrastructure-and-extensions/infrastructure-providers.mdx
@@ -14,10 +14,17 @@ The current infrastructure providers are:
 
 * [Bare metal provider](https://github.com/siderolabs/omni-infra-provider-bare-metal)
 * [KubeVirt provider](https://github.com/siderolabs/omni-infra-provider-kubevirt)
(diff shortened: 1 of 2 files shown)


public/omni/infrastructure-and-extensions/infrastructure-providers.mdx
```diff
@@ -14,10 +14,14 @@ The current infrastructure providers are:
 
 * [Bare metal provider](https://github.com/siderolabs/omni-infra-provider-bare-metal)
 * [KubeVirt provider](https://github.com/siderolabs/omni-infra-provider-kubevirt)
+* [Proxmox provider](https://github.com/siderolabs/omni-infra-provider-proxmox)
+* [Oxide provider](https://github.com/oxidecomputer/omni-infra-provider-oxide)
```
rothgar: Can you split this out into it's own list like "3rd party providers" just so people know it's not created or maintained by us

Issue #94, opened 2024-02-11
QEMU dev environment unable to bring up NTP because of failure to reach 8.8.8.8 DNS


dsseng:
## Bug Report
QEMU test cluster ran from [the guide](https://www.talos.dev/v1.6/advanced/developing-talos/#running-talos-cluster) fails because of inability to reach the 8.8.8.8 DNS resolver. 8.8.8.8 can be pinged from the host thus I believe it's a VM config issue or a compatibility problem.

### Description
```
validating CIDR and reserving IPs
generating PKI and tokens
creating "/tmp/tl-dev-home/.talos/cni/bin"
creating "/tmp/tl-dev-home/.talos/cni/cache"
creating "/tmp/tl-dev-home/.talos/cni/conf.d"
downloading CNI bundle from "https://github.com/siderolabs/talos/releases/download/v1.7.0-alpha.0/talosctl-cni-bundle-amd64.tar.gz" to "/tmp/tl-dev-home/.talos/cni/bin"
creating state directory in "/tmp/tl-dev-home/.talos/clusters/talos-default"
creating network talos-default
creating load balancer
creating dhcpd
creating controlplane nodes
creating worker nodes
waiting for API
^CSignal received, aborting, press Ctrl+C once again to abort immediately...
bootstrap error: 3 error(s) occurred:
        rpc error: code = DeadlineExceeded desc = context deadline exceeded
        rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 10.5.0.2:50000: connect: no route to host"
        context canceled
```

```
HOME=/tmp/tl-dev-home sudo --preserve-env=HOME _out/talosctl-linux-amd64 cluster create \                                                                                           [±main ✓]
    --provisioner=qemu \
    --registry-mirror docker.io=http://172.20.0.1:5000 \
    --registry-mirror registry.k8s.io=http://172.20.0.1:5001  \
    --registry-mirror gcr.io=http://172.20.0.1:5003 \
    --registry-mirror ghcr.io=http://172.20.0.1:5004 \
    --registry-mirror 127.0.0.1:5005=http://172.20.0.1:5005 \
    --install-image=127.0.0.1:5005/siderolabs/installer:v1.7.0-alpha.0-19-g5324d3916 \
    --controlplanes 1 \
    --workers 2 \
    --with-bootloader=false \
    --extra-uefi-search-paths "/tmp"
```

[...]

PR #222, opened 2025-12-12
docs: add info about ntp server overrides in omni

### LINKED ISSUE #156: Omni - document how to override NTP servers, etc
Have installed a Talos bare-metal machine on a server, but it struggles to reach Omni due to not able to reach a public NTP server.
I can override in the config, but when I generate the ISO in Omni, will the workaround be to manually patch it after with correct NTP or can it be "built-in" somehow?

aaah. Yes I think you can pass this as kernel arg, which will then be baked into the image. one sec
[9:42](https://talossystems.slack.com/archives/C09PDPU76GN/p1762958525899519?thread_ts=1762958281.814929&cid=C09PDPU76GN)
https://docs.siderolabs.com/talos/v1.11/reference/kernel#available-talos-specific-parameters
the ip argument is what you are looking for

This ⏫ flow should be clearly documented in Omni.
"How to override NTP servers or other network settings."

Iheanacho-ai:
fixes #156 #94


public/omni/cluster-management/override-ntp-servers.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+For example, to override the default NTP server (`time.cloudflare.com`) with `172.235.60.8`, you can specify only the NTP field and leave the rest empty:
+
+```bash
+ip=:::::::::172.235.60.8
+```
+
+You can then supply this `ip` value as a kernel parameter during boot to override the NTP server early in the startup process. To do this:
+
+1. In the Omni UI, select **Download Installation Media**.
```
rothgar: Can we also add a comment on how to do this via the CLI
```
omni download iso --extra-kernel-args "ip=:::::::"
```

rothgar (approved):

PR #247, opened 2025-12-18
docs: add airgapped discovey note

### LINKED ISSUE #248: Discovery access for Airgapped environments
Air gapped documentation does not mention the requirements for either disabling the discovery service, switching it to the K8s based discovery, or allowing access to https://discovery.talos.dev/  .  

Note that access to the discovery service is not essential, unless you are enabling KubeSpan..

Iheanacho-ai:
fixes #248

### DIFF
--- a/public/talos/v1.11/configure-your-talos-cluster/system-configuration/discovery.mdx
+++ b/public/talos/v1.11/configure-your-talos-cluster/system-configuration/discovery.mdx
@@ -48,7 +48,16 @@ cluster:
 
 Disabling all registries effectively disables member discovery.
 
-> Note: An enabled discovery service is required for [KubeSpan](../../networking/kubespan) to function correctly.
+<Note>
+Talos can operate normally with discovery disabled, and this does not affect day-to-day cluster operation.
+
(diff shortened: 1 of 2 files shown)


rothgar:
Omni actually has the discovery service built in so if they're using Omni they just need to check this box during cluster create.
image

the problem with discovery service is for air gapped environments that _don't_ use Omni

public/omni/infrastructure-and-extensions/install-airgapped-omni.mdx
```diff
@@ -417,7 +417,28 @@ Guides on creating a cluster on Omni can be found here:
...
+<Note>
+By default, Talos uses the public discovery service at:
+
+```
+https://discovery.talos.dev
+```
+
+In air-gapped or restricted environments, you must take one of the following actions:
+
+- Disable cluster discovery (`cluster.discovery.enabled: false`)
```
smira: there is no point in doing that actually, as Talos doesn't break if discovery.talos.dev is unavailable, it might be disabled to avoid log messages, but there will be no functional change

rothgar (approved):

PR #275, opened 2026-01-08
fix: fixed docs to run image-factory properly

kevintijssen:
I have changed the docs to make sure people run in image-factory  properly.
* Pinned to v0.9.0
* Fixed command to run image-factory with upstream
* Added and changes chapters

### DIFF
--- a/public/omni/infrastructure-and-extensions/self-hosted/deploy-image-factory-on-prem.mdx
+++ b/public/omni/infrastructure-and-extensions/self-hosted/deploy-image-factory-on-prem.mdx
@@ -34,6 +34,9 @@ INTERNAL_REG=127.0.0.1:5000
 
 ### Image Cache Signing Key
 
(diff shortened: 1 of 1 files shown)


public/omni/infrastructure-and-extensions/self-hosted/deploy-image-factory-on-prem.mdx
```diff
@@ -183,7 +191,7 @@ docker run -p 8080:8080 -d \
   --name image-factory \
   -v $PWD/signing-key.key:/signing-key.key:ro \
   -v $PWD/cosign.pub:/cosign.pub:ro \
-  ghcr.io/siderolabs/image-factory \
+  ghcr.io/siderolabs/image-factory:v0.9.0 \
```
rothgar: we should probably add this to the `snippets/custom-variables.mdx` so it's easier to change in the future.

PR #294, opened 2026-01-20
docs: add multi-doc patch

### LINKED ISSUE #202: Document multi-doc patch management
Call out how they get applied and used during maintenance mode and after a machine config is applied.

siderolabs/talos#12264

Iheanacho-ai:
fixes #202


public/talos/v1.12/configure-your-talos-cluster/system-configuration/patching.mdx
```diff
@@ -37,14 +37,6 @@ There are some special rules:
   - `cluster.apiServer.auditPolicy` value is replaced on merge
   - `ExtensionServiceConfig.configFiles` section is merged matching on `mountPath` (replacing `content` if matches)
 
-When patching a [multi-document machine configuration](../../reference/configuration/overview), following rules apply:
```
smira: I don't think we should push this down the document. Multi-doc config is now the default/common case, so we should just say this without "when", as it's always multi-doc.

PR #368, opened 2026-02-11
chore: add bannner

### LINKED ISSUE #323: conditional render a banner that indicates the newest version

Iheanacho-ai:
fixes #323

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -11,6 +11,10 @@ colors:
   light: "#FF5788"
   dark: "#FB326E"
 
+banner:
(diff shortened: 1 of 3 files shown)


common.yaml
```diff
@@ -11,6 +11,10 @@ colors:
   light: "#FF5788"
   dark: "#FB326E"
 
+banner:
```
rothgar: We can't use this function for the banner because if we ever want to create a banner (eg for kubecon or taloscon event) it would hide the version warning banner.

PR #396, opened 2026-02-24
docs: add limitation on argocd and multi-clusters

### LINKED ISSUE #386: manage multi-cluseters in Omni with aRGOCD
```
We have set up service accounts for the clusters using the omnictl command to create the kubeconfig and token for the accounts.
Can you please confirm for me:
1. Is there a way to view thse accounts in the clusters or onmi config? I cant see it in the ServiceAccounts on either.
2. Is this account locked to one cluster only?

I ask, as we are configuring ArgoCD to manage the clusters, and it doesnt seem to play nicely with two clusters under the same URL. But we are seeing a strange issue where the token created for the melbourne cluster seems to still sync to sydney in some cases, so we are troubleshooting what is happening with the accounts. I am going to confirm with another engineer that we are defintely using the separate tokens properly, but:
3. Do you have other customers that use Argo in the Omni world with multiple clusters?
4. Is there an option to have an alias, parameter, or something for the clusters that can be added to the URL to differentiate between the clusters?We replied:
1. Try omnictl servcieaccount list. ServiceAccounts have no dedicated resource.
2. Please help me understand this better: If you create a service account, it does not show up in the UI, neither listed by the above command? This is most certainly a bug, if you refer to an omni service account. However, I can't reproduce this, so we need further feedback. ServiceAccounts should definitively show up. ServiceAccounts are not limited to one cluster, as far as I know. They are ServiceAccounts for omni.
3+4: Not only customers - we do, in fact. There is a main issue with argocd in combination with omni-managed clusters (and any cluster that puts his kubeapi behind a proxy) here: the server value is the actual main key argocd uses, not it's alias (the name part on the cluster secret resource as described here). If omni now manages multiple of these clusters, they will (from an argocd viewpoint) share the same kubeapi URL. You can solve this in two ways: 1. run a reverse proxy, that manages a per-cluster unique kubeapi endpoint URL and simply forwards them to the omni kubernetes proxy. The second way is run one argocd instance in each target cluster.
In omni: Currently no. You can create a feature request for this, if you like.3+4 is what Noel refers to.
I am not sure we should add extensive docs on the subject, as it is a shortcoming of argocd and there has been a related issue for years that will probably be never addressed, as the related code is very deeply rooted in argocd. The problem of argocd is not specifically tied to omni, but applies to all k8s clusters, that have their kubeapi behind some sort of reverse proxy. As argocd internally uses the server URL as primary argocd Application destination identifier, two clusters provisioned by omni will appear to argocd as the same cluster, even if their name field is different in the related credentials file, compared to their server (kubeapi endpoint) field.
Possible solutions are:

run argocd separately in each cluster
run Ingress/reverse proxies providing a unique URL per kubeapi endpoint, forwarding to the omni kubeapi proxy.

```

Iheanacho-ai:
fixes #386

### DIFF
--- a/public/omni/troubleshooting/faqs.mdx
+++ b/public/omni/troubleshooting/faqs.mdx
@@ -32,3 +32,18 @@ Please see documents:
 * [Deploy Image Factory on-prem](../self-hosted/run-image-factory-on-prem.mdx)
 * [Deploy Omni on-prem](../self-hosted/run-omni-on-prem)
 
+## Why do multiple Omni clusters appear as the same cluster in ArgoCD?
+
+When using Omni-managed clusters behind the Omni kubeapi proxy, ArgoCD may treat multiple clusters as the same cluster if they share the same kubeapi server URL.
+
+This happens because ArgoCD identifies clusters primarily by their `server` URL, not by their cluster name or credentials. Since Omni exposes clusters through a shared kubeapi proxy endpoint, different clusters can appear identical from ArgoCD’s perspective.
+
+This is a limitation of ArgoCD and is not specific to Omni.
+
+To ensure clusters are treated separately, each cluster must have a unique `server` value in ArgoCD. You can do this by:
+
+- Appending a unique query parameter to the server URL (for example: `?cluster=cluster-a`)
+- Exposing each cluster through a distinct API endpoint (via DNS, reverse proxy, or ingress)
+- Running a separate ArgoCD instance per cluster
+
+As long as the `server` URL is different, ArgoCD will treat the clusters as separate clusters.


public/omni/troubleshooting/faqs.mdx
```diff
@@ -32,3 +32,17 @@ Please see documents:
...
+When using Omni-managed clusters behind the Omni kubeapi proxy, ArgoCD may treat multiple clusters as the same cluster if they share the same kubeapi server URL.
+
+This happens because ArgoCD identifies clusters primarily by their server URL, not by their cluster name or credentials. Since Omni exposes clusters through a shared kubeapi proxy endpoint, different clusters can appear identical from ArgoCD’s perspective.
+
+This is a limitation of ArgoCD and is not specific to Omni.
+
+To work around this limitation and ensure clusters are treated separately in ArgoCD, use one of the following approaches:
+
+- Run a separate ArgoCD instance in each target cluster
+- Configure ingress or reverse proxy routing so each cluster is exposed through a unique kubeapi endpoint URL
```
rothgar: Omni in the SaaS doesn't allow them to configure this.

@TimJones Is there another way to configure a single Argo for multiple Omni clusters?

PR #397, opened 2026-02-25
docs: add spegel

Iheanacho-ai:


public/talos/v1.10/configure-your-talos-cluster/images-container-runtime/spegel.mdx
```diff
@@ -0,0 +1,259 @@
...
+
+As the nodes reboot, they will temporarily become unavailable. Monitor their status and wait until they return to the `Ready` state:
+
+```bash
+kubectl get nodes --watch
+```
+
+Proceed once all nodes report `Ready`.
+
+### 1.4: Verify the configuration
```
rothgar: I don't think this step should be necessary and it won't work with Omni

PR #492, opened 2026-04-15
docs: clarify samllabelrules

### LINKED ISSUE #456: Dated documentation for SAMLLabelRules
[Documentation is dated for SAMLLabelRules](https://docs.siderolabs.com/omni/security-and-authentication/using-saml-with-omni/auto-assign-roles-to-saml-users)

It still recommends assignroleonregistration vs updated assignrole and fails 
`Error: failed to apply resources from file "omni/manifests/manifests.yaml": failed to update resource 'xxx' 'SAMLLabelRules.omni.sidero.dev': failed to update resource 'xxx' 'SAMLLabelRules.omni.sidero.dev': rpc error: code = InvalidArgument desc = failed to validate: 1 error occurred:
	* assignroleonregistration is deprecated, please use assignrole instead`

Also missing new `updateoneachlogin` in documentation

Iheanacho-ai:
fixes #456

### DIFF
--- a/public/omni/security-and-authentication/using-saml-with-omni/auto-assign-roles-to-saml-users.mdx
+++ b/public/omni/security-and-authentication/using-saml-with-omni/auto-assign-roles-to-saml-users.mdx
@@ -3,49 +3,87 @@ title: Auto-assign roles to SAML users
 description: Automatically assign Omni roles to new SAML users based on their SAML attribute labels.
 ---
 
-This guide shows you how to configure your Omni instance so that new users logging in with SAML authentication are automatically assigned to a role based on their SAML role attributes.
+When SAML authentication is enabled, Omni can automatically assign roles to new users based on their SAML attributes. To configure this, use SAMLLabelRule resources to map SAML attribute labels to Omni roles. If no rules are configured, [roles must be assigned manually](./add-a-user-to-omni-with-saml-enabled) by an administrator.
 
-Create the file `assign-operator-to-engineers-label.yaml` for the `SAMLLabelRule` resource, with the following content:
+## Prerequisites
 
-Copy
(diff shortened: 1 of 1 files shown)


public/omni/security-and-authentication/using-saml-with-omni/auto-assign-roles-to-saml-users.mdx
```diff
@@ -3,49 +3,61 @@ title: Auto-assign roles to SAML users
...
+## Prerequisites
 
-Copy
+- SAML authentication must be configured on your Omni instance. See [Authentication and Authorization](../authentication-and-authorization) for setup instructions.
+- `omnictl` must be installed and configured with Admin access to your Omni instance.
 
-```yaml
+## How SAML label rules work
+
+When a new SAML user logs in to Omni, Omni reads their SAML attributes and creates corresponding labels on the user resource. For example, a user with the SAML attribute `Role: engineers` will be labeled `saml.omni.sidero.dev/role/engineers` in Omni.
```
Unix4ever: Default known attributes for SAML are:

```go
	knownFields := map[string]string{
		// Microsoft
		"http://schemas.microsoft.com/ws/2008/06/identity/claims/role": "role",
		// Google
		"Role": "role",
		// Samltest
		"role": "role",
	}
```

So `Role: engineers` works.
But that really depends on the SAML identity provider. If it reports the role as something else, Omni should be configured with `--auth-saml-label-rules='{"custom-role-key": "role"}'` to map them.
Same thing should be done if the user wants to map any additional attributes coming from SAML to the user labels.

Not sure if we have the docs for that part. If we do, we should have the link here.

It should be clear in the docs that the first step should be to make the labels to appear on the users automatically and then they can also automatically map them to the Omni roles.

Unix4ever (approved):

PR #523, opened 2026-04-30
docs: add the omni api

Iheanacho-ai:


omni-api-gen/Makefile
```diff
@@ -0,0 +1,26 @@
+BINARY := omni-api-gen
```
rothgar: Usually the repo only has a top-level Makefile. This can get confusing with nested Makefile in directories. I suggest putting only basic targets in the top level

Unix4ever (commented):
We should probably use [kres](https://github.com/siderolabs/kres) here.
Then it will get rid of the extra Makefile.

omni-api-gen/Makefile
utkuozdemir: +1 from me

utkuozdemir:
> We should probably use [kres](https://github.com/siderolabs/kres) here. Then it will get rid of the extra Makefile.

kres-ify'ing this project is a bigger endeavor though, I'd leave it out of this PR.

omni-api-gen/main.go
```diff
@@ -0,0 +1,709 @@
...
+	"os"
+	"regexp"
+	"strings"
+)
+
+// protoFiles lists every .proto file to include, in display order.
+var protoFiles = []string{
+	"management/management.proto",
+	"oidc/oidc.proto",
+	"resources/resources.proto",
```
utkuozdemir: I'd say we should explicitly exclude this (resources.proto), we don't want to promote this as external API.

But also, I think, all the other ones should be discovered, not hardcoded. The generator should recursively search for the proto files in `./client/api/omni/` and include them, except there's an exclusion rule.

@Unix4ever wdyt?

utkuozdemir:
Good start! A few suggestions on the structure and framing:

1. The page could use a short introduction section before the table of contents, modeled on the [Omni Configuration](/omni/reference/omni-configuration) reference. A paragraph saying Omni has a gRPC API, schemas live at https://github.com/siderolabs/omni/tree/main/client/api, and that's what `protoc` consumers point at to generate stubs.

2. Worth saying somewhere that `omnictl` is usually what people want. It wraps these same APIs at a higher level and handles auth. Going direct over gRPC is for integrations and custom tooling.

3. The client example at https://github.com/siderolabs/omni/blob/main/client/pkg/client/example_test.go is probably the single most useful thing to link from the intro.

4. The biggest thing: I'd suggest splitting the page into two top-level sections, since Omni's API really has two distinct surfaces, and right now they're all flattened into a single list of proto files. Conceptually:

   - **Resource state API.** This is the generic COSI state API for managing resources (the same pattern Talos uses internally). The state API itself is defined upstream at https://github.com/cosi-project/specification/tree/main/proto/v1alpha1. Under this section, we'd document the `*Spec` messages from Omni's protos and the types they reference, since those are the payloads that flow through the state API. It might be worth a sentence or two explaining what the COSI state API is, since most readers won't have seen it before.
   - **Other APIs.** The purpose-built gRPC services: the management API, OIDC, and so on. This is roughly what's already in the doc today, just minus the resource specs (which move to the section above) and minus `resources/resources.proto` per my other comment.

   Splitting it this way matches how the codebase is organized and gives readers a much clearer mental model of which API does what.

Let me know if any of this is unclear and I'm happy to expand.


@Unix4ever @oguzkilcan do you think organizing it like I suggested in item 4 make sense?

omni-api-gen/main.go
Unix4ever: yes, that's only for the frontend

Unix4ever:
> Good start! A few suggestions on the structure and framing:
> 
> 1. The page could use a short introduction section before the table of contents, modeled on the [Omni Configuration](/omni/reference/omni-configuration) reference. A paragraph saying Omni has a gRPC API, schemas live at [siderolabs/omni@`main`/client/api](https://github.com/siderolabs/omni/tree/main/client/api), and that's what `protoc` consumers point at to generate stubs.
> 2. Worth saying somewhere that `omnictl` is usually what people want. It wraps these same APIs at a higher level and handles auth. Going direct over gRPC is for integrations and custom tooling.
> 3. The client example at [siderolabs/omni@`main`/client/pkg/client/example_test.go](https://github.com/siderolabs/omni/blob/main/client/pkg/client/example_test.go) is probably the single most useful thing to link from the intro.
> 4. The biggest thing: I'd suggest splitting the page into two top-level sections, since Omni's API really has two distinct surfaces, and right now they're all flattened into a single list of proto files. Conceptually:
>    
>    * **Resource state API.** This is the generic COSI state API for managing resources (the same pattern Talos uses internally). The state API itself is defined upstream at [cosi-project/specification@`main`/proto/v1alpha1](https://github.com/cosi-project/specification/tree/main/proto/v1alpha1). Under this section, we'd document the `*Spec` messages from Omni's protos and the types they reference, since those are the payloads that flow through the state API. It might be worth a sentence or two explaining what the COSI state API is, since most readers won't have seen it before.
>    * **Other APIs.** The purpose-built gRPC services: the management API, OIDC, and so on. This is roughly what's already in the doc today, just minus the resource specs (which move to the section above) and minus `resources/resources.proto` per my other comment.
>    
>    Splitting it this way matches how the codebase is organized and gives readers a much clearer mental model of which API does what.
> 
> Let me know if any of this is unclear and I'm happy to expand.
> 
> @Unix4ever @oguzkilcan do you think organizing it like I suggested in item 4 make sense?

Thanks @utkuozdemir for the great input.

To properly work with the Omni resource API we will need more things to be described:

5. We will need to describe how to work with Authentication: how to use service account keys in the Omni client. And yes, you can refer to this file https://github.com/siderolabs/omni/tree/main/client/api
6. We will need to describe how to work with the resources, their lifecycle: Create, Update, UpdateWithConflicts, Teardown, Destroy, TeardownAndDestroy. What's the difference between Teardown and Destroy calls, how UpdateWithConflicts works. Best practices.
7. We can have good descriptions written for the resources in the `.proto` files. Also it would be nice to have the access level near each resource which we generate. (Maybe we don't need to generate docs for the resources which are not readable by the user).

utkuozdemir:
Discussed it with @Unix4ever and @Iheanacho-ai  and we decided to put this on hold for now, because:
1. Without implementing the complicated/custom Omni auth with request payload signing, API is not usable. Implementing auth is not trivial.
2. Conversion of the resources' wrapped proto specs is another non-trivial challenge for API consumers.

`omnictl` and Go client library handles above concerns.

So we should first discuss and decide:
1. Who cares about an API reference and how much, and what do they actually expect (maybe Go client lib or embedding omnictl is fine for them).
2. What we should advertise / commit to as the Omni API.
3. Should we change anything on Omni and or build new things (e.g., adapters) or write additional documents (e.g., Omni Auth specification / how to unmarshal resources etc.) to address these complexities.

public/omni/reference/use-the-omni-api.mdx
```diff
@@ -0,0 +1,52 @@
+---
+title: The Omni API
```
rothgar: nit: Because this is under the reference/ folder people might assume to have the same information as talos/reference. We may want to rename this to something like "Integrate with the Omni API" or focus on how to consume it.

rothgar (approved):

utkuozdemir (approved):
I think this is a good start. The only thing I noticed is, we probably need to improve our example_test.go in Omni repo, with regular resource operations.

Edit: created https://github.com/siderolabs/omni/issues/2804 to track that.

public/omni/reference/build-with-the-omni-api.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+This is commonly used for observability integrations, automation pipelines, and systems that need to react to provisioning or cluster health events.
+
+The following resources are commonly used for status monitoring:
+
+| Resource | Description |
+|---|---|
+| `ClusterStatus` | Overall cluster health and lifecycle phase |
+| `MachineStatus` | Per-machine connectivity and provisioning status |
+| `ClusterMachineStatus` | Per-machine status within a specific cluster |
```
Unix4ever, line 48: Can also add `MachineStatusMetrics`, `ClusterStatusMetrics` to this list.

Unix4ever (approved):

PR #531, opened 2026-05-05
docs: add doc on stalled upgrade

Iheanacho-ai:


public/omni/cluster-management/upgrading-clusters.mdx
```diff
@@ -99,3 +99,50 @@ omnictl cluster machine lock <machine-id>
...
+
+### If a node appears stuck during an upgrade
+
+If a node is not progressing, wait before taking any action. Talos Linux performs several operations during an upgrade, including applying the new image, rebooting, and rejoining the cluster, and some steps can take longer than expected depending on hardware.
+
+If a node remains stuck after several minutes:
+
+1. Check the machine status in the Omni UI. A status of **Rebooting** or **Installing** indicates the upgrade is still in progress.
+
+2. If you have access to a serial or VNC console for the machine, use it to inspect the machine state directly. This is especially useful when Talos is unresponsive, since Talos does not allow SSH access by design.
```
Unix4ever: Serial/VNC console might be only needed if the machine is broken to the extent that it doesn't report the logs via SideroLink.
So I would suggest to first check machine logs in the Omni UI or using `omnictl machine-logs` command.

Unix4ever (approved):

PR #542, opened 2026-05-07
docs: document exposing a workload on multiple host ports

utkuozdemir:
The workload service proxy now accepts a comma-separated list of host port entries, optionally pairing each host port with a specific service port by number or name. Each entry produces its own exposed service with its own URL in Omni.

The label, prefix, and icon annotations gain per-host-port suffixed variants that override the unsuffixed one for a given host port.

Shipping in Omni 1.8. See siderolabs/omni#2771 and siderolabs/kube-service-exposer#18.


utkuozdemir:
We should hold this until Omni 1.8 lands.

Iheanacho-ai (approved):

PR #552, opened 2026-05-11
docs: add argocd and flux docs

### LINKED ISSUE #534: Document Omni manifest sync examples with ArgoCD and Flux
We have an example for talos https://docs.siderolabs.com/kubernetes-guides/advanced-guides/inlinemanifests#example-use-case-install-a-gitops-controller-with-extramanifests and we reference gitops approach for omni cluster management https://docs.siderolabs.com/omni/cluster-management/omni-terraform#the-recommended-approach-gitops-or-ci/cd

We should probably have some easy to follow copy/paste examples for deploying each one to bootstrap clusters and managing workloads.

Iheanacho-ai:
fixes #534

### DIFF
--- a/public/kubernetes-guides/advanced-guides/deploy-argocd.mdx
+++ b/public/kubernetes-guides/advanced-guides/deploy-argocd.mdx
@@ -1,0 +1,178 @@
+---
+title: "Deploy Argo CD"
+description: "In this guide you will learn how to deploy Argo CD on Talos Linux using extraManifests, or on Omni-managed clusters using manifest sync."
+---
+
+import { k8s_release } from '/snippets/custom-variables.mdx';
+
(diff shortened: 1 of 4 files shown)


public/kubernetes-guides/advanced-guides/deploy-argocd.mdx
```diff
@@ -0,0 +1,147 @@
...
+## Prerequisites
+
+Before you begin, ensure you have the following:
+
+- Talos 1.3 or later.
+- `kubectl` configured to access your cluster.
+- `talosctl` installed and configured, if deploying on Talos without Omni.
+- `omnictl` installed and configured, if deploying using Omni. See [Install and configure omnictl](../../omni/getting-started/install-and-configure-omnictl).
+
+## Installation on Talos
```
rothgar, line 21: Should these be tabs since it's one or the other?

Iheanacho-ai: not really, it would be better as tabs if they converged at one point, or diverged from one point.

These are completely different paths. So as long as I clearly state that these are different paths. It works.

rothgar (approved):

PR #602, opened 2026-06-12
docs: add information on how config patches work

### LINKED ISSUE #584: Docs: make the Omni cluster customization model evident (RFD 36)
This is the umbrella for the documentation work proposed in **RFD 36 — Making the cluster customization model evident**.

### Background

Omni built a powerful, scoped, weighted configuration-patch mechanism on top of Talos's strategic-merge primitive — but never documented the *model* (or under-documented it, at least): what a patch is, how scopes compose, which patch wins on conflict, and how patches relate to cluster templates. Today the patches page is a 17-line click-by-click how-to with no conceptual framing, and the authority-boundary page is partly wrong. New users can't converge on the correct mental model without reading the Talos source or filing a ticket.

This epic covers five docs changes that, together, teach the model where people already look.

### Key references

- RFD (full proposal): https://www.notion.so/siderolabs/RFD-36-Making-the-cluster-customization-model-evident-36fb1211badf812b86c8f4df99aba16f
- Conceptual model (source of truth — read this first): `docs/conceptual-model-cluster-customization.md`
- Figma designs (for the in-product views that these docs will stay consistent with once they're built): https://www.figma.com/design/pA5o8rTLwchmUwfyHWjygd/RFD-36

### ⚠️ The precedence rule — under correction (updated 2026-06-16)

**Do not use the rule that previously lived here.** The original wording — *"scope first, then weight within scope; the more specific scope wins, machine over machine-set over cluster"* — is **wrong**. It was caught during review of the docs page in #602 and corrected by engineering (Artem / @Unix4ever).

What's confirmed so far:

- There are **four** patch scopes, not three: `Machine`, `Cluster`, `MachineSet`, `ClusterMachine`.
- Priority is **low → high: `Machine` → `Cluster` → `MachineSet` → `ClusterMachine`.** A *standalone* `Machine` patch is the **lowest** (a cluster overrides it); a per-machine patch inside a cluster (`ClusterMachine`) is the **highest**. "More specific always wins" only half-holds.
- Within a single scope, patches are ordered **alphabetically by the padded-int ID prefix** (e.g. `000`, `001`, `400`). That prefix is the "weight."
- **Weight is set by a dedicated control, not the patch Name.** The Name is just a human-readable annotation (duplicates allowed). Do not tell users to encode weight in the name.

**Open question (blocking final wording):** does the everyday "pick a specific machine in a cluster and add a patch" map to `ClusterMachine` (highest)? Asked of Artem on #602 (L114). Once answered, the corrected canonical wording gets finalized here and folded into the conceptual model and RFD 36.

**Until then:** don't state a precedence rule in any customer-facing page, and hold the precedence/merge-order section of the "How configuration works in Omni" page pending this.

### Sub-issues

- [ ] New "How configuration works in Omni" page (the canonical explanation)
- [ ] Rewrite the Configuration Patches page to teach before the click-steps
- [ ] New "Where do I make this change?" decision guide
- [ ] Correct the authority-boundary ("Talos Config Overrides") page
- [ ] New "Configuration across upgrades" page

### Notes for the writer

- The two Omni authoring paths are **Configuration Patches** and **Cluster Templates** — and under the hood both produce the same patch resource. `talosctl` direct-to-machine is a third, Talos-native path that Omni reconciles away (a footgun for Omni-managed nodes); mention it only as an anti-pattern.
- "Talos Config Overrides" is **not** an authoring path — it's the list of fields Omni reserves. See the corrected-authority-boundary sub-issue.
- Weight has near-zero presence in the product today, but it **is** set via a dedicated control — **not** the patch Name (the Name is just an annotation; see the precedence section above). The docs should be the place this finally gets explained, once the open question is resolved.

Iheanacho-ai:
fixes #584

### DIFF
--- a/public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
+++ b/public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
@@ -1,0 +1,251 @@
+---
+title: How configuration works in Omni
+description: A conceptual overview of how Omni builds the configuration that runs on each Talos machine, what sources contribute, what you can change and how, and what happens when things silently fail.
+---
+
+When Omni manages a Talos machine, the configuration running on that machine is not written from one place.
+It is assembled from multiple sources, some you control, some Omni controls, and some exist entirely outside of Omni.
+
+Understanding where each piece of configuration comes from is the foundation for reasoning about cluster behavior, debugging unexpected outcomes, and knowing where to go when you need to make a change.
+
+## Six sources of configuration
+
+Every running Talos machine has configuration that was shaped by up to six distinct sources:
+
+1. **Patches**: Scoped YAML overrides authored through the Omni UI or cluster templates.
+2. **ExtensionsConfigurations**: Controls which system extensions, kernel modules, and firmware are baked into the boot image.
+3. **KernelArgs**: Controls the user-editable portion of the kernel command line.
+4. **KubernetesManifestGroups**: The list of Kubernetes manifests applied to the cluster, defined in a cluster template.
+5. **Platform or cloud-init injection**: Configuration written onto the machine by the platform before Talos starts.
+6. **Talos defaults**: Built-in behaviors that take effect when nothing else has been configured.
+
+The first four are inside Omni's authoring surface. The last two exist outside it, but they still affect what runs on the machine.
+
+### Patches
+
+[Patches](../reference/cluster-templates#patches) are the primary mechanism for customizing what runs on a machine. They are scoped YAML overrides that Omni merges into the final Talos machine configuration.
+
+Patches cover the bulk of what you would want to configure: networking, kubelet settings, registry mirrors, sysctls, time sync, KubeSpan, API server extra args, and more.
+
+You can author patches in two ways:
+
+- **Configuration Patches**: Created individually through the Omni UI or `omnictl`. This is the imperative path, suited for one-off changes.
+- **Cluster Templates**: A declarative multi-document YAML file that describes the entire desired state of a cluster, including patches at any scope, synced via `omnictl cluster template sync`. This is the path for GitOps workflows. See [Cluster Templates](../reference/cluster-templates) for more information.
+
+Both produce the same underlying `ConfigPatch` resource, go through the same controller, and are merged the same way. The choice between them is a workflow decision, not a mechanism decision.
+
+### ExtensionsConfigurations
+
+`ExtensionsConfigurations` is the resource you use to control which system extensions, kernel modules, and firmware are baked into a machine's boot image. It can be set at cluster, machine set, or machine level, through the Omni UI or via a cluster template. Omni resolves whichever `ExtensionsConfigurations` applies to a given machine into a per-machine `MachineExtensions` resource, which reflects the final result rather than something you edit directly.
+
+Changing extensions produces a new image. The machine performs a standard upgrade cycle to switch to the new image.
+
+### KernelArgs
+
+`KernelArgs` is a dedicated Omni resource for the user-editable portion of the kernel command line, for example, static IP configuration at boot time. Like extensions, kernel args have their own resource and are not set through patches.
+
+Omni manages the kernel arugments for SideroLink connectivity. That portion is set automatically for any installation media created by Omni and is not user-editable.
+
+### KubernetesManifestGroups
+
+`KubernetesManifestGroups` is the list of Kubernetes manifests being applied to the cluster. Unlike machine configuration, which is defined per machine, `KubernetesManifestGroups` is defined for the whole cluster only.
+
+Manifests are defined under `kubernetes.manifests` in a cluster template. Each manifest entry has a sync mode, `one-time` applies it once after bootstrap, `full` continuously reconciles it.
+
+### Platform or cloud-init injection
+
+On cloud and virtualized platforms, the provider writes configuration onto the machine before Talos has fully started, hostnames, network settings, metadata, through the `talos.config` kernel argument or cloud-init user-data. Talos does not merge configuration from multiple sources at once: a machine runs on either the full configuration injected via `apply-config`, or the configuration loaded from `talos.config`/cloud-init, not a combination of both. Omni does not currently act on this layer, so changes made here exist outside Omni's authoring surface.
+
+### Talos defaults
+
+Talos has built-in default behaviors that take effect when you have not explicitly configured something. DHCP assigning a hostname, default DNS resolution, default routing. Some of these can be overridden with a patch. Some cannot be turned off at all. None of them appear in any patch you have written, they are simply the baseline Talos starts from.
+
+<Note>
+When managing Talos Linux without Omni, `talosctl apply-config` and `talosctl edit machineconfig` are the primary authoring tools, they write directly to the node. In Omni-managed clusters, direct writes via `talosctl` are **blocked at the API layer**. Omni does not allow them to reach the machine. All configuration changes must go through Omni, using patches, `ExtensionsConfigurations`, or `KernelArgs` as appropriate.
+</Note>
+
+## Scope and precedence of how these configuration sources effect changes
+
+Patches and `ExtensionsConfigurations` can each be set at more than one level for the same machine. When that happens, scope determines which setting wins. The other configuration sources don't use scope or weight in this way: `KernelArgs` and `KubernetesManifestGroups` are each authored at a single level, and platform injection and Talos defaults sit outside Omni's authoring surface entirely.
+
+### Scope and precedence of patches
+
+Every patch targets exactly one scope. The scope determines which machines the patch applies to, and how long the patch persists.
+
+- **Machine**: Applies to one specific machine, independent of cluster membership. Created from the **Machines** UI page by selecting **Target: Machine**, or manually via a `ConfigPatch` resource labeled only with `omni.sidero.dev/machine: <uuid>`. A Machine patch persists until the machine is removed from Omni entirely, and is not applied while the machine is in maintenance mode.
+- **Cluster**: Applies to every machine in the cluster. Created from the cluster's **Config Patches** UI page by selecting **Cluster** as the Patch Target. Use for settings that should be uniform everywhere: registry mirrors, cluster DNS, KubeSpan.
+- **MachineSet**: Applies to every machine in a specific machine set, such as Control Planes or a named Workers group. Created from the cluster's Config Patches page by selecting the relevant machine set as the Patch Target. Use for role-specific configuration.
+- **ClusterMachine**: Applies to one specific machine within a cluster. Created from the cluster's Config Patches page by selecting an individual Node, or by defining a `patches` field in the `kind: Machine` block in a cluster template. A Cluster Machine patch is deleted when the cluster is deleted.
+
+A given machine can be affected by multiple scopes at once. When patches at different scopes set the same field, precedence determines which one wins.
+
+The most common scoping mistake is writing a cluster scoped patch for something that is logically per-machine. Hostname is the canonical example: the schema accepts it at cluster scope, but setting it there applies the same hostname to every machine in the cluster.
+
+#### Precedence: scope first
+
+When multiple patches set the same field, the more specific scope always wins. The priority order, from lowest to highest, is: Machine -> Cluster -> MachineSet -> ClusterMachine.
+
+Merging proceeds from lowest priority to highest:
+
+1. Machine scoped patches are applied first.
+2. Cluster scoped patches are applied next.
+3. MachineSet scoped patches are applied next.
+4. ClusterMachine scoped patches are applied last, and win on any conflicts.
+
+**Example.** A MachineSet patch sets `kubelet.maxPods = 200` across all workers. A ClusterMachine scoped patch on one specific worker sets `kubelet.maxPods = 100`. That worker ends up with `maxPods = 100`. The ClusterMachine scoped patch wins because it is highest in the priority order, regardless of what was set at MachineSet scope.
+
+#### Precedence: weight within a scope
+
+When two patches at the **same scope** set the same field, weight determines which one wins. Weight is a numeric value visible as a prefix in the patch ID — the `500-` in `500-2bf57a3a-cde5-45b7-8586-57ebd8235950`. The default weight is `500`.
+
+A higher weight means that patch is applied later and wins on conflicts within its scope.
+
+**Example, continued.** Suppose a second MachineSet patch also sets `kubelet.maxPods = 300`, with weight `700`. Both patches are at MachineSet scope. Weight `700` is higher, so `maxPods = 300` wins at the MachineSet level. But the ClusterMachine scoped patch still wins overall, because scope precedence is always resolved before weight. The final value for that specific worker is still `100`.
+
+The rule: **scope first, then weight within scope**. Weight never crosses scopes.
+
+Weight is controlled differently depending on how you author patches. In the UI, there is a dedicated **Weight field** in the patch editor. The default is `500` and valid values range from `100` to `900`. The patch ID is then generated automatically as `<weight>-<uuid>`.
+
+In cluster templates, weight is determined by the order of patches in the YAML array, starting at `400` and incrementing by one for each subsequent patch.
+
+### Scope and precedence of ExtensionsConfigurations
+
+An `ExtensionsConfigurations` resource is scoped by the labels attached to it. A cluster label alone applies it to the whole cluster. Adding a machine set label narrows it to that machine set. Adding a cluster machine label narrows it to that one machine.
+
+When more than one `ExtensionsConfigurations` resource applies to the same machine, the most specific one wins. This does not merge field by field — the extension list at the most specific level fully replaces the list from any less specific level, rather than combining with it.
+
+### How precedence works across the six configuration sources
+
+The six configuration sources, patches, `ExtensionsConfigurations`, `KernelArgs`, `KubernetesManifestGroups`, platform injection, and Talos Linux defaults, do not all have the same precedence.
+
+Patches have the highest priority and override values from any other source. Platform or cloud-init injection comes next. Its values can be overridden by patches, but they take precedence over Talos defaults. Talos defaults provide the baseline configuration and are applied only when a value has not been defined elsewhere.
+
+`ExtensionsConfigurations`, `KernelArgs`, and `KubernetesManifestGroups` operate outside this precedence hierarchy. They control different aspects of the system and do not overlap with patches, platform injection, Talos defaults, or with each other.
+
+## How configuration changes across Talos Linux versions
+
+The Talos machine configuration schema evolves over time. Fields may move to new configuration documents between releases.
+
+Talos maintains backward compatibility for patches across `1.x` releases, so patches that worked before an upgrade will continue to apply afterward. Talos default behavior for a cluster stays pinned to the Talos version it was created with, and Omni generates base machine configuration against that version going forward. This default behavior doesn't shift on its own — newer behavior introduced in later Talos versions isn't adopted until the cluster is upgraded.
+
+The hostname field provides a useful example of how field migrations are handled:
+
+- In older Talos releases, the hostname was configured using `machine.network.hostname`.
+
+- Starting in Talos 1.12, hostname configuration moved to a dedicated `HostnameConfig` document.
+
+When upgrading an existing cluster, the original field path continues to work because Talos preserves backward compatibility. However, when creating a new cluster on Talos 1.12 or later, a default `HostnameConfig` document is already present. In that case, patches that reference the old field path must be updated to use the new configuration format.
+
+Reviewing the [changelog](../../changelog) before an upgrade will surface any fields that have moved and any default behavior changes you'd be adopting.
+
+## What Omni controls vs. what you control
+
+Not every field in the Talos machine configuration is available to you. Omni partitions the configuration into three categories based on who authors what.
+
+### Reserved by Omni
+
+These fields are managed by Omni and cannot be set by a user patch. Patch attempts against them are either rejected with a `MergeDenied` error or silently stripped during reconciliation.
+
+Reserved fields include:
+
+- Cluster identity: `cluster.id`, `cluster.clusterName`
+- Cluster secrets and tokens: `cluster.secret`, `cluster.token`, bootstrap tokens
+- Cluster-level CAs and signing certs: `cluster.ca`, `cluster.aggregatorCA`, `cluster.serviceAccount`
+- Per-machine identity: `machine.ca`, `machine.token`
+- Specific derived fields like `cluster.controllerManager.extraArgs.service-account-key-file`
+
+For the authoritative list, see [Talos Config Overrides](../cluster-management/talos-config-overrides).
+
+### Managed through dedicated Omni resources
+
+Some fields are part of the Talos machine config schema, but Omni expects you to manage them through their Omni resources rather than through patches. A patch targeting these fields will not be rejected outright, but the behavior will not be what you expect.
+
+| Field | Use instead |
+|---|---|
+| `machine.install.extensions` | `ExtensionsConfigurations` resource or Cluster Templates (recommended) |
+| `machine.install.extraKernelArgs` (user-editable portion) | `KernelArgs` resource |
+
+### Configured through patches
+
+Everything else, the bulk of `machine.*` and `cluster.*` outside the reserved and dedicated-resource fields, is configured through patches. This includes:
+
+<Note>Note that some fields, particularly under `machine.network.*`, have moved to separate multi-doc configuration documents in newer Talos versions.</Note>
+
+| Field area | Notes |
+|---|---|
+| `machine.network.*` | Schema is mid-migration across Talos versions. See [How configuration changes across Talos versions](#how-configuration-changes-across-talos-versions). |
+| `machine.kubelet.*` | Kubelet flags and config. |
+| `machine.sysctls`, `machine.kernelModules` | Kernel tuning. |
+| `machine.time.*` | NTP servers. See [Override NTP Servers](../cluster-management/override-ntp-servers). |
+| `machine.registries.*` | Registry mirrors and pull credentials. |
+| `cluster.discovery.*`, `cluster.kubespan.*` | Cluster networking and discovery. |
+| `cluster.apiServer.extraArgs.*` | Some keys (e.g., `audit-log-*`) are accepted by the patch validator but silently dropped by Talos. |
+| `cluster.controllerManager.extraArgs.*` | Except `service-account-key-file`, which is reserved. |
+| `cluster.scheduler.extraArgs.*`, `cluster.proxy.*` | Standard cluster-level config. |
+| `cluster.inlineManifests.*` | Manifest sync. |
+
+## Reference: where to make changes
+
+Use this table to find where to make a specific change without having to work through the full model.
+
+### Cluster-wide settings
+
+Applies to every machine in the cluster.
+
+| If you want to... | Where | Notes |
+|---|---|---|
+| Set NTP servers | Configuration Patches or Cluster Templates | — |
+| Set cluster DNS | Configuration Patches or Cluster Templates | — |
+| Add a registry mirror | Configuration Patches or Cluster Templates | — |
+| Configure KubeSpan | Configuration Patches or Cluster Templates | — |
+| Sync a Kubernetes manifest | Cluster Templates | Backed by the `KubernetesManifestGroup` resource. Not available as a configuration patch; manifests can currently only be viewed in the UI, not set there. See [Sync Kubernetes Manifests](../cluster-management/sync-kubernetes-manifests). |
+| Add a custom CA certificate | Configuration Patches or Cluster Templates | Use a `TrustedRootsConfig` document in the patch. This appends to the system trusted certificate store and is distinct from the cluster CA which is reserved by Omni. |
+
+### Role-specific settings
+
+Applies to a machine set, for example, all control-plane nodes or all GPU workers.
+
+| If you want to... | Where | Notes |
+|---|---|---|
+| Set sysctls or kernel modules per role | Configuration Patches or Cluster Templates | Use cluster scope if uniform across all roles; machine set scope if role-specific. |
+| Add a system extension to a specific role | `ExtensionsConfigurations` resource | Not a patch. Changing extensions re-images the machine. |
+
+### Per-machine settings
+
+Applies to one specific machine.
+
+| If you want to... | Where | Notes |
+|---|---|---|
+| Set a static IP | `KernelArgs` resource, embedded machine config, or platform user-data | A machine needs network connectivity to reach Omni before it can receive a patch, so static IP configuration is set ahead of that connection rather than through one. |
+| Set the hostname | Configuration Patches or Cluster Templates | Use `HostnameConfig` syntax on Talos 1.12+. |
+| Set a kernel argument | `KernelArgs` resource | Not a patch. |
+
+### Managed automatically by Omni, not user-editable
+
+These fields are owned entirely by Omni. Attempting to change them through a patch will either return an error or have no effect.
+
+| If you want to... | Where | Notes |
+|---|---|---|
+| Change a hidden SideroLink kernel argument | — | Set automatically by Omni. Not user-editable. |
+| Change the cluster name or cluster ID | — | Reserved by Omni. Cannot be changed after cluster creation. |
+
+### Patches that silently won't do what you expect
+
+These patches are accepted without error but do not produce the expected result. If your patch appears to apply but nothing changes on the machine, check whether the field falls into one of these cases.
+
+| If you want to... | What happens | Notes |
+|---|---|---|
+| Set `audit-log-*` keys under `apiServer.extraArgs` | Accepted by the patch validator; silently dropped by Talos | The patch goes through cleanly but does nothing. |
+| Set `controllerManager.extraArgs.service-account-key-file` | Patch returns `MergeDenied` | Reserved by Omni. |
+| Apply config directly via `talosctl` | Blocked at the Omni API layer | In Omni-managed clusters, use Configuration Patches or Cluster Templates instead. |
+
+## Troubleshooting
+
+Each failure mode in Omni configuration has a distinct shape. Use the symptoms to identify which layer is in play.
+
+| Symptom | Likely cause | Where to look |
+|---|---|---|
+| Patch applied, value missing from merged config | Field is reserved by Omni, or patch is not scoped to this machine | [Talos Config Overrides](../cluster-management/talos-config-overrides); check patch scope |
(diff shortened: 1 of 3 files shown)


public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+### Patches
+
+[Patches](../reference/cluster-templates#patches) are the primary mechanism for customizing what runs on a machine. They are scoped YAML overrides that Omni merges into the final Talos machine configuration.
+
+Patches cover the bulk of what you would want to configure: networking, kubelet settings, registry mirrors, sysctls, time sync, KubeSpan, API server extra args, and more.
+
+You can author patches in two ways:
+
+- **Configuration Patches** — created individually through the Omni UI or `omnictl`. This is the imperative path, suited for one-off changes.
+- **Cluster Templates** — a declarative multi-document YAML file that describes the entire desired state of a cluster, including patches at any scope, synced via `omnictl cluster template sync`. This is the path for GitOps workflows.
```
rothgar: Don't we have a docs page explaining cluster templates? Maybe link to it

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+Both produce the same underlying `ConfigPatch` resource, go through the same controller, and are merged the same way. The choice between them is a workflow decision, not a mechanism decision.
+
+### MachineExtensions
+
+`MachineExtensions` is a dedicated Omni resource for which system extensions, kernel modules, and firmware are baked into the machine's boot image. You do not add extensions through a patch. You set them through the `MachineExtensions` resource in the Omni UI or via a cluster template.
+
+Changing extensions produces a new image. The machine re-images on the next boot.
```
rothgar:
```suggestion
Changing extensions produces a new image. The machine performs a standard upgrade cycle to switch to the new image.
```

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+### KernelArgs
+
+`KernelArgs` is a dedicated Omni resource for the user-editable portion of the kernel command line, for example, static IP configuration at boot time. Like extensions, kernel args have their own resource and are not set through patches.
+
+Omni also manages a hidden portion of the kernel command line for SideroLink connectivity. That portion is set automatically and is not user-editable.
```
rothgar:
```suggestion
Omni manages the kernel arugments for SideroLink connectivity. That portion is set automatically for any installation media created by Omni and is not user-editable.
```

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+### Platform or cloud-init injection
+
+On most cloud and bare-metal platforms, the provider writes configuration onto the machine before Talos has fully started, hostnames, network settings, metadata. Talos reconciles on top of whatever the platform wrote. Omni does not see this layer and cannot control it.
```
rothgar: Bare metal usually **doesn't** do this. It's typically only a cloud/virtualized option, but even on-prem virtualized don't often use it.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+### Talos defaults
+
+Talos has built-in default behaviors that take effect when you have not explicitly configured something. DHCP assigning a hostname, default DNS resolution, default routing. Some of these can be overridden with a patch. Some cannot be turned off at all. None of them appear in any patch you have written, they are simply the baseline Talos starts from.
+
+<Note>
+In standalone Talos, `talosctl apply-config` and `talosctl edit machineconfig` are the primary authoring tools, they write directly to the node. In Omni-managed clusters, direct writes via `talosctl` are **blocked at the API layer**. Omni does not allow them to reach the machine. All configuration changes must go through Omni, using patches, `MachineExtensions`, or `KernelArgs` as appropriate.
```
rothgar: I would say "When managing Talos Linux without Omni..." IMO "standalone Talos" makes it sound like single node clusters.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+- Cluster secrets and tokens: `cluster.secret`, `cluster.token`, bootstrap tokens
+- Cluster-level CAs and signing certs: `cluster.ca`, `cluster.aggregatorCA`, `cluster.serviceAccount`
+- Per-machine identity: `machine.ca`, `machine.token`
+- Specific derived fields like `cluster.controllerManager.extraArgs.service-account-key-file`
+
+For the authoritative list, see [Talos Config Overrides](../cluster-management/talos-config-overrides).
+
+### Managed through dedicated Omni resources
+
+Some fields are part of the Talos machine config schema, but Omni expects you to manage them through their own dedicated resources rather than through patches. A patch targeting these fields will not be rejected outright, but the behavior will not be what you expect.
```
rothgar:
```suggestion
Some fields are part of the Talos machine config schema, but Omni expects you to manage them through their Omni resources rather than through patches. A patch targeting these fields will not be rejected outright, but the behavior will not be what you expect.
```

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+| Field | Use instead |
+|---|---|
+| `machine.install.extensions` | `MachineExtensions` resource |
+| `machine.install.extraKernelArgs` (user-editable portion) | `KernelArgs` resource |
+
+### Configured through patches
+
+Everything else, the bulk of `machine.*` and `cluster.*` outside the reserved and dedicated-resource fields, is configured through patches. This includes:
+
+| Field area | Notes |
```
rothgar, line 174: You'll want to update this to show configuration that is part of a multi-doc config (eg network)

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+- **Machine Set** — applies to every machine in a specific machine set, such as all control-plane nodes or all GPU workers. Use for role-specific configuration.
+- **Machine** — applies to one specific machine. Use for per-machine settings like a static IP or a unique hostname.
+
+A given machine is affected by all three scopes simultaneously. Cluster-scoped patches, machine-set-scoped patches targeting its machine set, and patches scoped directly to it all apply at the same time.
+
+The most common scoping mistake is writing a cluster-scoped patch for something that is logically per-machine. Hostname is the canonical example: the schema accepts it at cluster scope, but setting it there applies the same hostname to every machine in the cluster.
+
+### Precedence: scope first
+
+When multiple patches set the same field, the more specific scope always wins. Machine scope overrides machine-set scope; machine-set scope overrides cluster scope.
```
rothgar: You probably need to expand this to include what happens with default values, cloud-init config, embedded machine config, kernel args, etc.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+When two patches at the **same scope** set the same field, weight determines which one wins. Weight is a numeric value visible as a prefix in the patch ID — the `500-` in `500-2bf57a3a-cde5-45b7-8586-57ebd8235950`. The default weight is `500`.
+
+A higher weight means that patch is applied later and wins on conflicts within its scope.
+
+**Example, continued.** Suppose a second machine-set patch also sets `kubelet.maxPods = 300`, with weight `700`. Both patches are at machine-set scope. Weight `700` is higher, so `maxPods = 300` wins at the machine-set level. But the machine-scoped patch still wins overall, because scope precedence is always resolved before weight. The final value for that specific worker is still `100`.
+
+The rule: **scope first, then weight within scope**. Weight never crosses scopes.
+
+To set a non-default weight, include a numeric prefix when naming the patch. There is no labeled weight field in the UI — the prefix in the name is the only control.
```
rothgar: Is this just set by the "Name" field in the UI?

e.g.

my-patch       <- gets 500-my-patch weight
600-my-patch <- gets 600-my-patch weight

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+## How configuration changes across Talos versions
+
+The Talos machine config schema is not static. Fields move, syntax changes, and semantics shift between releases. A patch that works correctly today can silently stop working after a Talos upgrade.
```
rothgar: We should be very careful with wording here because we try extremely hard to keep backwards compatibility for upgrades. The patch should never stop working, but default settings on Talos machines may change which may influence how a configuration works.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+The hostname field is the clearest example:
+
+- In **Talos 1.10 and 1.11**, hostname was set at `machine.network.hostname`. A patch like `machine.network.hostname: edge-03` worked as expected.
+- In **Talos 1.12+**, hostname configuration moved to a separate `HostnameConfig` document. The old `machine.network.hostname` field is still accepted, it parses cleanly and applies cleanly, but it no longer does anything. The hostname falls back to DHCP with no error or warning.
```
rothgar: I don't think this is true. You can still use the older-style hostname as long as you're doing an upgrade. If you apply it to a new config you'd have to remove the default HostNameConfig multi-doc file

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+This pattern is not limited to hostname. The Talos team is currently breaking the entire `machine.network` block into separate documents, [`HostnameConfig`](../../talos/v1.13/reference/configuration/network/hostnameconfig), [`BondConfig`](../../talos/v1.13/reference/configuration/network/bondconfig), [`BridgeConfig`](../../talos/v1.13/reference/configuration/network/bridgeconfig), [`LinkConfig`](../../talos/v1.13/reference/configuration/network/linkconfig), [`DHCPConfig`](../../talos/v1.13/reference/configuration/network/dhcpv4config), and others. As each piece lands in a new Talos release, patches written against the old syntax silently no-op on the affected fields. There is no diff, no warning in the Omni UI, and no migration tool.
```
rothgar: I would leave this paragraph out because it's constantly changing and Omni docs aren't versioned so depending on what version of Talos the user is on would impact what multi-doc config they have

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+Talos engineering aims for backward compatibility when fields move, and old syntax often continues to work through a transition window. But the policy is not guaranteed, and there is no surface inside the product that tells you whether your patch syntax is still effective after an upgrade.
```
rothgar: I _think_ we do guarantee backwards compatibility with talos 1.x releases.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+## Reference: where to make changes
+ 
+Use this table to find where to make a specific change without having to work through the full model.
+ 
+### Cluster-wide settings
+ 
+Applies to every machine in the cluster.
+ 
+| If you want to... | Where | Notes |
```
rothgar, line 195: You should probably add Custom CA certificates and static IP address to this list. They're common configurations.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+| Configure KubeSpan | Configuration Patches or Cluster Templates | |
+| Sync a Kubernetes manifest | Configuration Patches or Cluster Templates | See [Sync Kubernetes Manifests](../cluster-management/sync-kubernetes-manifests). |
+ 
+### Role-specific settings
+ 
+Applies to a machine set — for example, all control-plane nodes or all GPU workers.
+ 
+| If you want to... | Where | Notes |
+|---|---|---|
+| Set kubelet flags per role | Configuration Patches or Cluster Templates | One patch per machine set. |
```
rothgar: Why would someone set this? For registration labels?

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+| Set sysctls or kernel modules per role | Configuration Patches or Cluster Templates | Use cluster scope if uniform across all roles; machine set scope if role-specific. |
+| Add a system extension to a specific role | `MachineExtensions` resource | Not a patch. Changing extensions re-images the machine. |
+ 
+### Per-machine settings
+ 
+Applies to one specific machine.
+ 
+| If you want to... | Where | Notes |
+|---|---|---|
+| Set a static IP | Configuration Patches or Cluster Templates | Use `LinkConfig` syntax on Talos 1.12+. The older `machine.network.*` block silently no-ops on migrated fields. |
```
rothgar: They would actually need to do this via cloud-init, kernel args, or embedded machine config. Setting it as a patch won't let a machine connect to Omni if they don't have DHCP

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+| Set a kernel argument | `KernelArgs` resource | Not a patch. |
+ 
+### Managed automatically by Omni, not user-editable
+ 
+These fields are owned entirely by Omni. Attempting to change them through a patch will either return an error or have no effect.
+ 
+| If you want to... | Where | Notes |
+|---|---|---|
+| Change a hidden SideroLink kernel argument | — | Set automatically by Omni. Not user-editable. |
+| Rotate cluster CAs, signing certs, or cluster secrets | Omni certificate and secret management flows | Patches against these fields return `MergeDenied`. |
```
rothgar: They're allowed to rotate the CA via omnictl. They just can't set a static CA.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
sterlingkoch: Justin is correct and I got it wrong when I said this, @Iheanacho-ai.

sterlingkoch: Maybe rework it to be built around something like "A config created fresh on 1.12+ needs the new partial-doc syntax (eg. HostnameConfig)"

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
sterlingkoch: Agree.  The field-by-field detail belongs in the versioned Talos docs, not this Omni explanation page.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
sterlingkoch: Another spot where I think I led you astray, @Iheanacho-ai.  It should be pointed at cloud-init / kernel args / embedded machine config for first-boot static IP, and reserve patches / LinkConfig for changes to an already-connected machine.  And you should probably confirm that's actually correct with someone in eng.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
sterlingkoch: To expand on this, this page introduces 5 configuration sources at the top (L11-19), but the precedence section only resolves the three patch scopes. Readers will ask how patches relate to defaults, cloud-init, and kernel args in the final result.  A short "how the five sources compose" note would close the loop.  Even just stating that patches/MachineExtensions/KernelArgs are what you author, while platform injection and Talos defaults sit underneath and can be overridden would be helpful.  Doesn't need to be long.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+If you are upgrading Talos and you manage network configuration through patches, review the Talos release notes for that version and verify that any fields you depend on have not moved to a newer document type.
+
+## Debugging configuration issues
+ 
+Each failure mode in Omni configuration has a distinct shape. Use the symptoms to identify which layer is in play.
+ 
+| Symptom | Likely cause | Where to look |
+|---|---|---|
+| Patch applied, value missing from merged config | Field is reserved by Omni, or patch is not scoped to this machine | [Talos Config Overrides](/omni/cluster-management/talos-config-overrides); check patch scope |
```
sterlingkoch: Minor nit that we use relative link paths everywhere else on this page, but an absolute one here.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
sterlingkoch: The answer to Justin's question is "yes".  There's no separate weight field in the editor. 

Tt might be worth adding these exact examples to the doc, since it makes it instantly clear.

sterlingkoch (changes requested):

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
rothgar: it would make it clear that's how it works and also make it so someone will depend on it working that way

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+You can author patches in two ways:
+
+- **Configuration Patches** — created individually through the Omni UI or `omnictl`. This is the imperative path, suited for one-off changes.
+- **Cluster Templates** — a declarative multi-document YAML file that describes the entire desired state of a cluster, including patches at any scope, synced via `omnictl cluster template sync`. This is the path for GitOps workflows.
+
+Both produce the same underlying `ConfigPatch` resource, go through the same controller, and are merged the same way. The choice between them is a workflow decision, not a mechanism decision.
+
+### MachineExtensions
+
+`MachineExtensions` is a dedicated Omni resource for which system extensions, kernel modules, and firmware are baked into the machine's boot image. You do not add extensions through a patch. You set them through the `MachineExtensions` resource in the Omni UI or via a cluster template.
```
Unix4ever: User editable resource is `ExtensionsConfiguration`.
It can be defined on 3 levels: cluster, machine set, machine.

`MachineExtensions` also has this information but it's produced from `ExtensionsConfiguration` resource for each machine by Omni.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
Unix4ever: Omni sees this layer and can control, but doesn't do that as of now.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
Unix4ever:
```suggestion
In standalone Talos, `talosctl apply-config` and `talosctl edit machineconfig` are the primary authoring tools, they write directly to the node. In Omni-managed clusters, direct writes via `talosctl` are **blocked at the API layer**. Omni does not allow them to reach the machine. All configuration changes must go through Omni, using patches, `ExtensionsConfigurations`, or `KernelArgs` as appropriate.
```

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+For the authoritative list, see [Talos Config Overrides](../cluster-management/talos-config-overrides).
+
+### Managed through dedicated Omni resources
+
+Some fields are part of the Talos machine config schema, but Omni expects you to manage them through their own dedicated resources rather than through patches. A patch targeting these fields will not be rejected outright, but the behavior will not be what you expect.
+
+| Field | Use instead |
+|---|---|
+| `machine.install.extensions` | `MachineExtensions` resource |
```
Unix4ever:
```suggestion
| `machine.install.extensions` | `ExtensionsConfigurations` resource |
```

Cluster templates can define that too and that's an intended way to do it.
In the templates it's just the list of extensions you define in the cluster, machine set or machine resource.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+Since patches are the dominant authoring mechanism, it is worth understanding precisely how they compose.
+
+### Scopes
+
+Every patch targets exactly one scope. The scope determines which machines the patch applies to.
+
+- **Cluster** — applies to every machine in the cluster. Use for settings that should be uniform everywhere: registry mirrors, cluster DNS, KubeSpan.
+- **Machine Set** — applies to every machine in a specific machine set, such as all control-plane nodes or all GPU workers. Use for role-specific configuration.
+- **Machine** — applies to one specific machine. Use for per-machine settings like a static IP or a unique hostname.
```
Unix4ever: There are more scopes for the node level:
`ClusterMachine` and `Machine`.

- `ClusterMachine` config patch is deleted when you delete the cluster.
- `Machine` config patch stays until you remove the machine from Omni completely. But it's not applied in maintenance mode.

Unix4ever: The priority of the patches is the following (from low to high):

`Machine` -> `Cluster` -> `MachineSet` -> `ClusterMachine`.

Inside the group the order is alphabetical. That's why we weight the patches by padded ints (000, 001, 400).

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
Unix4ever: Name field in the UI doesn't matter for order. You can even have duplicates, as it's just an annotation added to a patch. Human readable representation of what that patch is doing.

In the UI you can't controller the patch id: it gets the weight you define and appends random UUID to the end of it.
In the cluster template the weight is calculated by the order of patches in the list you have in the YAML file (something like `400-<patch-name/file-name>`)

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
Unix4ever: Talos keeps backward compatibility for the patches. So they don't break silently.
If it does, then it's a bug.

Also Omni remembers the initial Talos version the cluster was created with, so it preserves the set of enabled features in the config for that cluster.
It's called version contract and it also adds more predictability to the patch generation.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
Unix4ever: Old style patches are rejected by Omni, but you can still use them with the new Talos.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+## Debugging configuration issues
+ 
+Each failure mode in Omni configuration has a distinct shape. Use the symptoms to identify which layer is in play.
+ 
+| Symptom | Likely cause | Where to look |
+|---|---|---|
+| Patch applied, value missing from merged config | Field is reserved by Omni, or patch is not scoped to this machine | [Talos Config Overrides](/omni/cluster-management/talos-config-overrides); check patch scope |
+| Patch applied, value in merged config, machine ignores it | Field is silently dropped by Talos (e.g. `audit-log-*` under `apiServer.extraArgs`), or a Talos default is overriding it | Check Talos docs for whether the specific key is supported |
+| Configuration broke after a Talos upgrade, no error surfaced | Field moved to a new document type; old syntax silently no-ops | Talos release notes for the version you upgraded to |
```
Unix4ever: this shouldn't happen

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+| Machine has a value you never set | Platform or cloud-init injection, or a Talos default (e.g. DHCP-assigned hostname) | Not visible in any patch list, exists outside Omni's authoring surface |
```
Unix4ever: Cloud init injection will be gone as soon as you allocate the machine into a cluster.

Talos doesn't merge the configs from these sources: it's either full Talos config injected by `apply-config` call or it's the config loaded from `talos.config` kernel arg or cloud-init user-data.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+ 
+Applies to every machine in the cluster.
+ 
+| If you want to... | Where | Notes |
+|---|---|---|
+| Set NTP servers | Configuration Patches or Cluster Templates | |
+| Set cluster DNS | Configuration Patches or Cluster Templates | |
+| Add a registry mirror | Configuration Patches or Cluster Templates | |
+| Configure KubeSpan | Configuration Patches or Cluster Templates | |
+| Sync a Kubernetes manifest | Configuration Patches or Cluster Templates | See [Sync Kubernetes Manifests](../cluster-management/sync-kubernetes-manifests). |
```
Unix4ever: Oh, btw `KubernetesManifestGroup` is used for that. You don't use Configuration Patches for that. CLI way is `Cluster Templates`, that's correct.

It's not possible to set manifests in the UI now. I doubt we should have it there.
You can view them only

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,225 @@
...
+ 
+### Role-specific settings
+ 
+Applies to a machine set — for example, all control-plane nodes or all GPU workers.
+ 
+| If you want to... | Where | Notes |
+|---|---|---|
+| Set kubelet flags per role | Configuration Patches or Cluster Templates | One patch per machine set. |
+| Set sysctls or kernel modules per role | Configuration Patches or Cluster Templates | Use cluster scope if uniform across all roles; machine set scope if role-specific. |
+| Add a system extension to a specific role | `MachineExtensions` resource | Not a patch. Changing extensions re-images the machine. |
```
Unix4ever: Wrong resource again and can be set on all levels: cluster, machine set, machine.

Unix4ever:
We need to add a note about `KubernetesManifestGroups` here. That's another resource type we have for configuring stuff.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
sterlingkoch: @Unix4ever - let me make sure I'm understanding the model right.

Priority low→high: Machine → Cluster → MachineSet → ClusterMachine, alphabetical by padded-int ID prefix within each group. So a standalone Machine patch is the lowest (below cluster), while a ClusterMachine patch is the highest.  Is that correct?

Follow-up question: when someone picks a specific machine from the Patch Target dropdown on a cluster and creates a patch there, is that a ClusterMachine patch (highest)? And is a Machine patch only what you get when you author against an unallocated machine that isn't in a cluster yet?

@Iheanacho-ai - If the above mapping holds, the everyday user-facing rule can stay close to "the more specific scope wins" (cluster → machine-set → the machine, in a cluster), with standalone Machine patches documented as a separate, lowest-priority baseline that a cluster overrides. If instead users routinely encounter both Machine and ClusterMachine as distinct choices for the same node, we need to teach the full four-scope ordering explicitly

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,238 @@
...
+
+Every running Talos machine has configuration that was shaped by up to five distinct sources:
+
+1. **Patches**: Scoped YAML overrides authored through the Omni UI or cluster templates.
+2. **ExtensionsConfigurations**: Controls which system extensions, kernel modules, and firmware are baked into the boot image.
+3. **KernelArgs**: Controls the user-editable portion of the kernel command line.
+4. **Platform or cloud-init injection**: Configuration written onto the machine by the platform before Talos starts.
+5. **Talos defaults**: Built-in behaviors that take effect when nothing else has been configured.
+
+The first three are inside Omni's authoring surface. The last two exist outside it, but they still affect what runs on the machine.
```
Unix4ever:
```suggestion

1. **Patches**: Scoped YAML overrides authored through the Omni UI or cluster templates.
2. **ExtensionsConfigurations**: Controls which system extensions, kernel modules, and firmware are baked into the boot image.
3. **KernelArgs**: Controls the user-editable portion of the kernel command line.
4. **KubernetesManifestGroups**: ...
5. **Platform or cloud-init injection**: Configuration written onto the machine by the platform before Talos starts.
6. **Talos defaults**: Built-in behaviors that take effect when nothing else has been configured.


The first four are inside Omni's authoring surface. The last two exist outside it, but they still affect what runs on the machine.
```

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,238 @@
...
+
+- **Machine**: Applies to one specific machine, independent of cluster membership. Created from the **Machines** UI page by selecting **Target: Machine**, or manually via a `ConfigPatch` resource labeled only with `omni.sidero.dev/machine: <uuid>`. A Machine patch persists until the machine is removed from Omni entirely, and is not applied while the machine is in maintenance mode.
+- **Cluster**: Applies to every machine in the cluster. Created from the cluster's **Config Patches** UI page by selecting **Cluster** as the Patch Target. Use for settings that should be uniform everywhere: registry mirrors, cluster DNS, KubeSpan.
+- **MachineSet**: Applies to every machine in a specific machine set, such as Control Planes or a named Workers group. Created from the cluster's Config Patches page by selecting the relevant machine set as the Patch Target. Use for role-specific configuration.
+- **ClusterMachine**: Applies to one specific machine within a cluster. Created from the cluster's Config Patches page by selecting an individual Node, or by defining a `kind: Machine` block in a cluster template. A Cluster Machine patch is deleted when the cluster is deleted.
```
Unix4ever:
```suggestion
- **ClusterMachine**: Applies to one specific machine within a cluster. Created from the cluster's Config Patches page by selecting an individual Node, or by defining `patches` field in the `kind: Machine` block in a cluster template. A Cluster Machine patch is deleted when the cluster is deleted.
```

Same is true for cluster template blocks on other levels:
```yaml
---
kind: Cluster
patches:
   - ...
---
kind: ControlPlane
patches:
   - ...
---
kind: Workers
patches:
   - ...
---
kind: Machine
patches:
   - ...
```

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,238 @@
...
+
+In cluster templates, weight is determined by the order of patches in the YAML array, starting at `400` and incrementing by one for each subsequent patch.
+
+### How precedence works across the five configuration sources
+
+The five configuration sources, patches, `ExtensionsConfigurations`, `KernelArgs`, platform injection, and Talos Linux defaults, do not all have the same precedence.
+
+Patches have the highest priority and override values from any other source. Platform or cloud-init injection comes next. Its values can be overridden by patches, but they take precedence over Talos defaults. Talos defaults provide the baseline configuration and are applied only when a value has not been defined elsewhere.
+
+`ExtensionsConfigurations` and `KernelArgs` operate outside this precedence hierarchy. They control different aspects of the system and do not overlap with patches, platform injection, Talos defaults, or with each other.
```
Unix4ever: Worth mentioning that `ExtensionsConfigurations` also have that hierarchy:
Cluster -> MachineSet -> ClusterMachine.

But it doesn't merge the extensions list, it overwrites the list fully with the next levels.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,238 @@
...
+
+## How configuration changes across Talos Linux versions
+
+The Talos machine configuration schema evolves over time. Fields may move to new configuration documents, and default behaviors can change between releases. 
+
+Talos maintains backward compatibility for patches across `1.x` releases, so patches that worked before an upgrade will continue to apply afterward. However, changes to Talos defaults can still affect the resulting machine configuration. Even when a patch is applied successfully, a change in default behavior may influence the final configuration that Talos generates.
```
Unix4ever: I'd say default behavior shouldn't change unless there's a bug.
Omni remembers the initial version of the Talos cluster, so if cluster was created for Talos 1.11, Omni will preserve enabled features for it: will disable new stuff that might break the default behavior.

When Omni generates the base config for Talos nodes it will use `1.11` as a version contact always.

public/omni/omni-cluster-setup/how-configuration-works-in-omni.mdx
```diff
@@ -0,0 +1,238 @@
...
+
+The hostname field provides a useful example of how field migrations are handled:
+
+- In older Talos releases, the hostname was configured using `machine.network.hostname`. 
+
+- In newer releases, hostname configuration was moved to a dedicated `HostnameConfig` document. 
+
+When upgrading an existing cluster, the original field path continues to work because Talos preserves backward compatibility. However, when creating a new cluster on a newer Talos version, a default `HostnameConfig` document is already present. In that case, patches that reference the old field path must be updated to use the new configuration format.
```
Unix4ever: @smira is that accurate to say that `HostnameConfig` config is already present for the config generated for Talos 1.13+?
I think Talos just falls back to the defaults.

smira: HostnameConfig is present since 1.12+ by default: https://github.com/siderolabs/talos/blob/fcd2ae0f9201e773f329600b219b4144b3e2d355/pkg/machinery/config/types/v1alpha1/testdata/stability/v1.12/base-controlplane.yaml#L91-L95

This is the place to check default machine configs for any version of Talos

Unix4ever (approved):

sterlingkoch (approved):

shanduur (approved):

PR #603, opened 2026-06-12
docs: clarify SecureBoot auto-enrollment is VM-only by default

mcanevet:
## What?

Clarify the bare-metal SecureBoot guide (v1.13 and v1.14) around key enrollment:

- **Flow diagram** relabeled `Setup Mode: Auto-enroll Keys` → `Setup Mode: Enroll Keys` — it implied keys are auto-enrolled unconditionally in setup mode.
- Added a note that **automatic enrollment only happens inside a VM** (the default `secure-boot-enroll` mode is `if-safe`), with the bare-metal alternatives.
- Reworded the *"if the UEFI firmware does not support automatic enrollment"* sentence, which framed it as a firmware capability when it's actually the `if-safe` policy.

The bare-metal `force` alternative is version-specific: v1.13 supports it via a custom SecureBoot **ISO**, while v1.14 supports it for disk images too (auto-enrollment for disk images merged in 1.14).

## Why?

The guide and the diagram read as if booting a SecureBoot image on bare-metal in setup mode enrolls the keys automatically. It doesn't: sd-boot's `if-safe` only auto-enrolls under virtualization. On real hardware the keys are enrolled via the manual `Enroll Secure Boot keys: auto` boot-menu entry, or by building an image with `secure-boot-enroll` set to `force`. The previous wording sends bare-metal users down a path that silently does nothing.

See siderolabs/talos#12568 for the discussion.

### DIFF
--- a/public/talos/v1.13/platform-specific-installations/bare-metal-platforms/secureboot.mdx
+++ b/public/talos/v1.13/platform-specific-installations/bare-metal-platforms/secureboot.mdx
@@ -42,7 +42,7 @@ The SecureBoot process follows a strict verification chain from UEFI firmware to
 flowchart TD
     UEFI[UEFI Firmware]
     KEYS{SecureBoot Keys<br/>Enrolled?}
-    SETUP[Setup Mode<br/>Auto-enroll Keys]
+    SETUP[Setup Mode<br/>Enroll Keys]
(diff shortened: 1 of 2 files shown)


Iheanacho-ai:
@smira what do you think of this

smira (approved):

PR #622, opened 2026-06-25
docs: update the update kubernetes docs

### LINKED ISSUE #563: Following the Manual upgrade guide rolls back kube-proxy
Path: /kubernetes-guides/advanced-guides/upgrading-kubernetes
[Upgrading Kubernetes - Manual Kubernetes Upgrade](https://docs.siderolabs.com/kubernetes-guides/advanced-guides/upgrading-kubernetes#manual-kubernetes-upgrade)

When I follow the steps in the guide to manually upgrade Kubernetes I get the following problem:
After upgrading `kube-apiserver`, `kube-controller-manager` and `kube-scheduler` I upgrade the `kube-proxy` DaemonSet as described in [the guide](https://docs.siderolabs.com/kubernetes-guides/advanced-guides/upgrading-kubernetes#proxy).
This works as expected, but when I move on to the next step ([Bootstrap manifests](https://docs.siderolabs.com/kubernetes-guides/advanced-guides/upgrading-kubernetes#bootstrap-manifests)), the manifests returned using the command in the docs includes a `kube-proxy` DaemonSet that contains both the old image version, and immutable config ( the `.spec.selector`  config), and applying this will naturally roll back the `kube-proxy` DaemonSet image version.
```
$talosctl get manifests -o yaml | yq eval-all '.spec | .[] | splitDoc' - > manifests.yaml
$kubectl diff -f manifests.yaml
The DaemonSet "kube-proxy" is invalid: spec.selector: Invalid value: {"matchLabels":{"k8s-app":"kube-proxy","tier":"node"}}: field is immutable
```
The reason i get the old image version is (of course) because I have `kube-proxy` defined in my `machineconfig` under `.cluster.proxy.image`.

I don't know if this is an error in the docs, or if this is an unusual way for my cluster to be configured, but for me, I would need to update the version in `machineconfig` before syncing the bootstrap manifests in addition to deleting the immutable fields from the `kube-proxy` DaemonSet to do a successful Kubernetes upgrade.

If relevant: My cluster was not initially created with the `talosctl gen config` command, but migrated from a `kubeadm` cluster following [the official guide](https://docs.siderolabs.com/talos/v1.13/advanced-guides/migrating-from-kubeadm).

```
$talosctl version
Client:
        Tag:         v1.13.0
        SHA:         b9e9c657
        Built:
        Go version:  go1.26.2
        OS/Arch:     linux/amd64
Server:
        NODE:        <IP>
        Tag:         v1.13.0
        SHA:         b9e9c657
        Built:
        Go version:  go1.26.2
        OS/Arch:     linux/amd64
        Enabled:     RBAC
```

Iheanacho-ai:
fixes #563

### DIFF
--- a/public/kubernetes-guides/advanced-guides/upgrading-kubernetes.mdx
+++ b/public/kubernetes-guides/advanced-guides/upgrading-kubernetes.mdx
@@ -148,13 +148,21 @@ See the [Reproducible Machine Configuration](../../talos/v1.10/configure-your-ta
 
 ## Manual Kubernetes upgrade
 
-Kubernetes can be upgraded manually by following the steps outlined below.
-They are equivalent to the steps performed by the `talosctl upgrade-k8s` command.
+Kubernetes can be upgraded manually by following the steps outlined below. They are equivalent to the steps performed by the `talosctl upgrade-k8s` command.
+
+Before running the commands below, export the `k8s_release` variable, replacing the placeholder with the target Kubernetes version:
+
+```bash
+export k8s_release=<target-k8s-version>
+```
+
+<Note>
+The `talosctl patch mc` examples use strategic-merge (YAML) patches, which work across supported Talos versions. From Talos v1.13 the machine configuration is a multi-document configuration, for which strategic-merge patches should be used rather than RFC 6902 (`op`/`path`/`value`) patches. A strategic-merge patch also creates the target key if it is not already present.
+</Note>
 
 ### Kubeconfig
 
-In order to edit the control plane, you need a working `kubectl` config.
-If you don't already have one, you can get one by running:
+In order to edit the control plane, you need a working `kubectl` config. If you don't already have one, you can get one by running:
 
 ```bash
 talosctl --nodes <control-plane-ip> kubeconfig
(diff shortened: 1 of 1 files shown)


public/kubernetes-guides/advanced-guides/upgrading-kubernetes.mdx
```diff
@@ -148,30 +148,37 @@ See the [Reproducible Machine Configuration](../../talos/v1.10/configure-your-ta
...
 
-Kubernetes can be upgraded manually by following the steps outlined below.
-They are equivalent to the steps performed by the `talosctl upgrade-k8s` command.
+Kubernetes can be upgraded manually by following the steps outlined below. They are equivalent to the steps performed by the `talosctl upgrade-k8s` command.
+
+The commands below reference `${k8s_release}` (the target Kubernetes version) and `${k8s_prev_release}` (the currently running version). Export these variables, replacing the placeholders with your values:
+
+```bash
+export k8s_release=<target-k8s-version>
+export k8s_prev_release=<current-k8s-version>
```
sterlingkoch: k8s_prev_release is exported but never used in the manual steps. The rewrite removed the old kube-proxy DaemonSet YAML that referenced the "from" image (${k8s_prev_release}). After this change, every command in the manual section uses only ${k8s_release}. So both the prose ("…and ${k8s_prev_release} (the currently running version)") on line 153 and the export k8s_prev_release=... on line 157 reference a variable that no longer appears in any command below. Suggest dropping k8s_prev_release from this block to avoid sending readers off to find a value they'll never use.

public/kubernetes-guides/advanced-guides/upgrading-kubernetes.mdx
```diff
@@ -211,31 +218,25 @@ spec:
...
 In this example, the new version is `5`.
-Wait for the new pod definition to propagate to the API server state (replace `talos-default-controlplane-1` with the node name):
-
-```bash
-kubectl get pod -n kube-system -l k8s-app=kube-apiserver --field-selector spec.nodeName=talos-default-controlplane-1 -o jsonpath='{.items[0].metadata.annotations.talos\.dev/config\-version}'
-5
-```
 
-Check that the pod is running:
+Wait for the new pod definition to propagate, then check that the `kube-apiserver` pods are running on the control plane nodes:
```
sterlingkoch: The new verification command can't actually confirm the upgrade propagated. Each section now says "Wait for the new pod definition to propagate, then check that the pods are running" followed by kubectl get pods -n kube-system -l k8s-app=… -o wide. But -o wide shows node/IP, not the container image — so it confirms pods are running, not that they're running the new version. The text it replaced verified the talos.dev/config-version annotation, which actually proved propagation. The old jsonpath was admittedly fragile, but we've lost the verification entirely. Suggest either restoring an image/version check, e.g.:

```
kubectl get pods -n kube-system -l k8s-app=kube-apiserver \
  -o jsonpath='{.items[*].spec.containers[*].image}'
or keeping a lightweight config-version check.
```


Same for line 330

sterlingkoch (changes requested):

sterlingkoch (approved):

PR #606, opened 2026-06-16
docs: update discovery service docs to multi-doc

majabojarska:
Updates documentation to reflect changes made for https://github.com/siderolabs/talos/issues/13243

Depends on (must not be merged before): https://github.com/siderolabs/talos/pull/13562

- Fixes `docs-convert` handling of multiple-paragraph `description` values in the frontmatter.
- We allow configuring multiple discovery service endpoints now, expressed
through multiple DiscoveryServiceConfig documents.
- `.cluster.discovery` is deprecated, but it's the only way to perform
  discovery through Kubernetes registry (which is also deprecated).
- Moves discovery service to the top, and kubernetes registry to the
  bottom, in the "Registries" section. The discovery service is the most
relevant nowadays.
- Removes the video walkthrough, as it instructs the user to use the
  deprecated `.cluster.discovery` block.
- Adds links to the discovery service page from the air-gapped config and KubeSpan doc pages, for the end user's convenience.

PR #626, opened 2026-06-26
docs(talos): add LVM provisioning and management docs

shanduur:
New LVM page: volume groups, logical volumes, raw volume backings, inspect/remove.
Cross-references from disk-management pages. Nav entry in talos-v1.14.yaml.

Signed-off-by: Mateusz Urbanek <mateusz.urbanek@siderolabs.com>

### DIFF
--- a/public/talos/v1.14/configure-your-talos-cluster/storage-and-disk-management/disk-management/user.mdx
+++ b/public/talos/v1.14/configure-your-talos-cluster/storage-and-disk-management/disk-management/user.mdx
@@ -129,6 +129,62 @@ Restrictions for `disk` volumes:
 * The provisioning block must not define size-related settings (`minSize`, `maxSize`, or `grow`).
   A full device implies a fixed size.
 
+#### User volumes on LVM logical volumes
+
+An LVM logical volume can back a user volume.
+Use `volumeType: disk` so Talos treats the logical volume as the block device for the user volume instead of trying to create another partition on it.
+
+First create the <a href={`../../../reference/configuration/storage/lvmvolumegroupconfig`}>LVMVolumeGroupConfig</a> and <a href={`../../../reference/configuration/storage/lvmlogicalvolumeconfig`}>LVMLogicalVolumeConfig</a>.
+For example, create a logical volume that uses the full volume group:
+
+```yaml
+apiVersion: v1alpha1
+kind: LVMLogicalVolumeConfig
+name: lv-user
+type: linear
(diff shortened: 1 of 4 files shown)


public/talos/v1.14/configure-your-talos-cluster/storage-and-disk-management/disk-management/user.mdx
```diff
@@ -129,6 +129,62 @@ Restrictions for `disk` volumes:
...
+```
+
+After the logical volume is created, find its stable device symlink.
+The logical volume appears as a device-mapper disk with symlinks such as `/dev/mapper/vg--pool-lv--user` or `/dev/disk/by-dm-name/vg--pool-lv--user`.
+
+```bash
+talosctl get disks -o yaml
+```
+
+Then create a `UserVolumeConfig` of type `disk` and match that symlink:
```
majabojarska: I'm thinking we could get this PR first, which regenerates the Talosctl reference https://github.com/siderolabs/docs/pull/606. It includes the reference for `UserVolumeConfig`, and then we could link to it directly from here.

majabojarska: That one's merged, so it should now be possible to link to reference.

smira (approved):
👍

public/talos/v1.14/configure-your-talos-cluster/storage-and-disk-management/disk-management/user.mdx
majabojarska: These `href` tags don't seem to not render properly in the generated site (`make preview`)
image

 I think we can just use the basic MD hyperlink syntax, but the URL needs to be fixed as well. We can link to the same-version talosctl reference without specifying the version in the path:
 
```diff
-Then create a <a href={`../../talos/${version}/reference/configuration/block/uservolumeconfig`}>UserVolumeConfig</a> of type `disk` and match that symlink:
+Then create a [UserVolumeConfig](../../../reference/configuration/block/uservolumeconfig) of type `disk` and match that symlink:
```

image

The site generator will resolve the appropriate absolute URL.

majabojarska: Same for the other occurrences: `LVMVolumeGroupConfig`, `LVMLogicalVolumeConfig`.

PR #638, opened 2026-07-10
docs: update the upgrade talos version make target

Iheanacho-ai:
What

Makes make upgrade-talos-version driven by an explicit tag instead of guessing the next version.

```
make upgrade-talos-version TAG=v1.14.0-beta.0   # or v1.14.0 for stable
```

Why

The old target guessed the version by bumping the minor, which overshot whenever the pins were already advanced during an alpha. It also never regenerated reference docs and didn't validate input.

Changes

- TAG is the source of truth: stage from the suffix, docs folder from the minor.
- Regenerates Talos reference docs automatically.
- nvidia values fall back to latest-stable instead of ending up empty.
- Validates the tag up front with two clear errors: malformed vs unpublished.
- GITHUB_TOKEN auto-fills from gh and is exported to recipes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,14 @@ TALOS_VERSION := v1.14
 VALE_IMAGE := jdkato/vale:latest
 VALE_CONFIG ?= .vale.ini
 
+# Auto-fill the GitHub token from `gh` when it isn't already set in the environment.
+# An exported/CI value wins; otherwise fall back to the local gh keychain. If gh is
+# unavailable this is simply empty, matching the previous behaviour.
+# `export` makes it visible to recipe subprocesses (e.g. the local `go run` targets),
+# not just to the containers that pass it explicitly with `-e`.
+GITHUB_TOKEN ?= $(shell gh auth token 2>/dev/null)
+export GITHUB_TOKEN
+
 # Default target
 .PHONY: help
 help: ## Show this help message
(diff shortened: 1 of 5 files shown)


public/talos/v1.14/reference/cli.mdx
```diff
@@ -134,7 +134,7 @@ talosctl cluster create dev [flags]
 
 ```
       --airgapped                                limit VM network access to the provisioning network only
-      --arch string                              cluster architecture (default "amd64")
+      --arch string                              cluster architecture (default "arm64")
```
sterlingkoch: I'm confused (and possibly wrong in my confusion) about this change.
I think this should actually be "amd64".  Same with line 355.
If I'm right, I'm guessing the culprit of regenerating the docs on an Apple Silicon (arm64) machine.

Iheanacho-ai: That's true, it turns out that value depended on which computer ran talosctl docs. I've fixed it and updated the make target logic so it stays consistent.

Makefile
```diff
@@ -331,29 +345,53 @@ changelog-local: ## Generate the changelog using local Go build
...
+		echo "  Pass a TAG whose talosctl image already exists."; \
+		exit 1; }
+endef
+
 .PHONY: upgrade-talos-version
-upgrade-talos-version: ## Upgrade docs to the next Talos minor version (fetches versions from GitHub)
+upgrade-talos-version: ## Upgrade Talos docs to a release tag: make upgrade-talos-version TAG=v1.14.0-beta.0
+	@test -n "$(TAG)" || { echo "Error: TAG is required, e.g. make upgrade-talos-version TAG=v1.14.0-beta.0"; exit 1; }
+	$(validate-tag)
 	docker run --rm -it -v $(PWD):/workspace -w /workspace \
```
sterlingkoch: Claude tells me that we should remove `-it` here since it's now fully `--tag` driven (non-interactive).
`-it` fails with "the input device is not a TTY" in any non-TTY context (ie. CI, output piped to a file). Since the interactivity is gone, drop `-it` entirely (or a `-t` at most).

sterlingkoch (approved):

PR #639, opened 2026-07-11
docs: clean out the nav

Iheanacho-ai:
Made sure every file in the folder is represented in its respective YAML file, and added a check to the docs CI that validates each file is included in its YAML.

Also, deleted duplicate `cli.yaml` files across versions 1.6 to 1.11


talos-v1.14.yaml
```diff
@@ -283,6 +302,8 @@ navigation:
                     - "configuration/network/resolverconfig.mdx"
                     - "configuration/network/routingruleconfig.mdx"
                     - "configuration/network/statichostconfig.mdx"
+                    - "configuration/runtime/sysctlconfig.mdx"
```
sterlingkoch: configuration/runtime/sysctlconfig.mdx and configuration/runtime/sysfsconfig.mdx are added inside the Network group. They're runtime/ configs, not network/ — they'll render under the wrong sidebar heading. 

Unless I'm missing something, these should be their own Runtime group (like the new Cluster, Kubernetes, Storage groups added in this same PR).

sterlingkoch (approved):

PR #644, opened 2026-07-15
docs: add the info on disabling selinux secureboot

### LINKED ISSUE #578: Documentation Update for Disabling SELinux specifically with SecureBoot
Hi,
When using secureboot, if I try to disable SELinux by setting `selinux=0` per the official documentation(1), then SELinux is not actually disabled. It still ends up in permissive mode. I suspect that this is because /proc/cmdline ends up with `... selinux=1 selinux=0 ...`. Which -- I thought was supposed to be 'last one wins', but that doesn't appear to be working the way I understood.

I had to instead add `-selinux` which worked.
Example:
```yaml
customization:
  extraKernelArgs:
    - -selinux
```

I wanted to open this because I could not find the info in any of the github issues or documentation, and it took me an embarrassingly long time to work it out (which I consider to mostly be an issue on my side between keyboard-and-chair).

- Talos version: 1.13.3
- Kubernetes version: 1.32
- Platform: linux / bare-metal with secureboot

Ref:
(1) https://docs.siderolabs.com/talos/v1.13/security/selinux#disabling-selinux

PS: Talos is great and y'all are amazing.

Iheanacho-ai:
fixes #578

### DIFF
--- a/public/talos/v1.13/security/selinux.mdx
+++ b/public/talos/v1.13/security/selinux.mdx
@@ -26,8 +26,14 @@ is not compatible with some of your workloads or plugins.
 
 SELinux also needs to be disabled if you are enabling AppArmor, because both security modules cannot be enabled at the same time.
 
-If you want to disable SELinux, you can do so by adding `selinux=0` to the kernel cmdline.
-This is most commonly done when creating the image via the configuration in the Image Factory.
+To disable SELinux, add `-selinux selinux=0` to the kernel cmdline. This is done when creating the image through the Image Factory configuration. In your Image Factory schematic:
+
+```yaml
+customization:
+  extraKernelArgs:
+    - -selinux
+    - selinux=0
+```
 
 ### Mode of operation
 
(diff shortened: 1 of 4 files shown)


public/talos/v1.13/security/selinux.mdx
```diff
@@ -29,6 +29,14 @@ SELinux also needs to be disabled if you are enabling AppArmor, because both sec
 If you want to disable SELinux, you can do so by adding `selinux=0` to the kernel cmdline.
 This is most commonly done when creating the image via the configuration in the Image Factory.
 
+On SecureBoot images, use `-selinux` instead. Talos includes `selinux=1` in the signed kernel command line by default, and SecureBoot prevents it from being overridden at runtime. Adding `selinux=0` leaves both arguments present (`selinux=1 selinux=0`), which leaves SELinux in permissive mode. To disable it, remove the default argument in your Image Factory schematic:
```
smira: this has nothing to do with SecureBoot, just amend the previous paragraph to say use `-selinux` && `selinux=0` as the extra kernel args to replace the default

sterlingkoch (approved):

PR #666, opened 2026-07-22
docs: chore rearrange all the make targets into one tools folder

Iheanacho-ai:


sterlingkoch (commented):
It looks like we deleted Claude.md.  Was that intentional?  Just wanted to make sure before approving.

Iheanacho-ai:
> It looks like we deleted Claude.md. Was that intentional? Just wanted to make sure before approving.

yes it was intentional

sterlingkoch (approved):

PR #667, opened 2026-07-23
docs: add note on infra provider

### LINKED ISSUE #661: Troubleshooting: removing machines owned by an infrastructure provider
The [troubleshooting page's "Removing stuck machines" section](https://docs.siderolabs.com/omni/troubleshooting/troubleshooting) covers omnictl delete link but doesn't cover machines created by an infrastructure provider, which behave differently.

* `omnictl delete machine fails with PermissionDenied` — only read access is permitted. Nothing tells you Machine is read-only and Link is the deletable resource. Worth stating explicitly.
* Provider-owned machines carry provider finalizers, so delete link alone leaves them half-deleted (visible in UI, gone from omnictl get machine) unless the provider is running to process the cleanup.
* Deleting the provider registration removes its machines. Recent provider versions implement proper cleanup on provider deletion (confirm current behavior with Utku/Oguz before writing, and note which provider versions it applies to).

Suggestion: Either extend the existing troubleshooting section or add a short "Removing provider-managed machines" subsection.

Iheanacho-ai:
fixes #661

### DIFF
--- a/public/omni/troubleshooting/troubleshooting.mdx
+++ b/public/omni/troubleshooting/troubleshooting.mdx
@@ -8,24 +8,36 @@ The sections are unordered and can be used to troubleshoot your specific situati
 
 ## Removing stuck machines
 
-Machines connected to Omni can get in states where Omni can no longer cleanly remove them. This may be because the machine is no longer available, the machine has errors, or other reasons.
+Machines connected to Omni can sometimes become stuck and cannot be removed cleanly. This may happen because the machine is no longer reachable, has entered an error state, or another cleanup operation failed.
 
-This may block the ability to delete a cluster as Omni tries to cleanly remove machines before deleting the cluster resources. Sometimes trying to force delete the machine from the Omni UI also does not remove the machine.
+Depending on how the machine was created, the removal process differs.
 
-To delete a machine from Omni via `omnictl` you will first need to identify what SideroLink it is using.
+### Machines managed directly by Omni
 
-Identify the UUID of the machine that is stuck deleting by copying it from the menu.
+For machines managed directly by Omni, first identify the SideroLink associated with the machine.
+
+Copy the machine UUID from the UI.
 
 ![Machine menu](./images/machine-menu.png)
 
-The machine UUID should also match a SideroLink UUID from
+Verify that the UUID matches a SideroLink:
 
 ```bash
 omnictl get links
 ```
 
-Once you verify one of the link UUIDs match you can delete the link (and machine) with:
+Delete the link:
 
 ```bash
 omnictl delete link $UUID
 ```
+
+Deleting the link also removes the associated machine.
+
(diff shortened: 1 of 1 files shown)


public/omni/troubleshooting/troubleshooting.mdx
```diff
@@ -8,24 +8,53 @@ The sections are unordered and can be used to troubleshoot your specific situati
...
 
-Identify the UUID of the machine that is stuck deleting by copying it from the menu.
+For machines managed directly by Omni, first identify the SideroLink associated with the machine.
+
+Copy the machine UUID from the UI.
 
 ![Machine menu](./images/machine-menu.png)
 
-The machine UUID should also match a SideroLink UUID from
+Verify that the UUID matches a SideroLink:
```
Unix4ever, line 23: UUID always matches SideroLink link resource ID.
We can drop this sentense.

public/omni/troubleshooting/troubleshooting.mdx
```diff
@@ -8,24 +8,53 @@ The sections are unordered and can be used to troubleshoot your specific situati
...
+Machines connected to Omni can sometimes become stuck and cannot be removed cleanly. This may happen because the machine is no longer reachable, has entered an error state, or another cleanup operation failed.
 
-This may block the ability to delete a cluster as Omni tries to cleanly remove machines before deleting the cluster resources. Sometimes trying to force delete the machine from the Omni UI also does not remove the machine.
+Depending on how the machine was created, the removal process differs.
 
-To delete a machine from Omni via `omnictl` you will first need to identify what SideroLink it is using.
+### Machines managed directly by Omni
 
-Identify the UUID of the machine that is stuck deleting by copying it from the menu.
+For machines managed directly by Omni, first identify the SideroLink associated with the machine.
```
Unix4ever, line 17: I would suggest using the old 
```suggestion
For machines managed directly by Omni, first identify the UUID of the machine that is stuck deleting.
```

public/omni/troubleshooting/troubleshooting.mdx
```diff
@@ -8,24 +8,53 @@ The sections are unordered and can be used to troubleshoot your specific situati
...
+```
+
+returns an error similar to:
+
+```text
+rpc error: code = PermissionDenied
+desc = only read access is permitted
+```
+
+This is expected behavior. These machines are read-only because their lifecycle is managed by the infrastructure provider.
```
Unix4ever: `omnictl delete machine` doesn't work for any machines. So this whole block is not correct.

The same `omnictl delete link` flow works for the infra provider managed machines. And it is ok to use that for the broken machines created by the infra providers. No leftovers should remain after that, but there is one thing to keep in mind: if the cluster requested the machine from the infra provider is running, deleted link will be immediately replaced by the new one.

sterlingkoch (approved):

PR #668, opened 2026-07-23
docs: fix argo and flux docs

### LINKED ISSUE #647: Issue on docs: after bootstrapping argocd, argocd pods exist but not in argocd namespace
Path: /kubernetes-guides/advanced-guides/deploy-argocd

Hey - I'm following this doc and fairly new to talos.   I almost but not quite got the result I was expecting.  Wondering what I did wrong.  I didn't want to use the patch method, rather I wanted to add the provided details from the doc to my controlplane.yaml so that I could bootstrap my cluster with argo in less steps in the event of some kind of cataclysm.

Goal: Install argocd at bootstrap time with argocd pods in the argocd namespace

Result: argocd pods are running, but not in the argocd namespace.  Argocd namespace does exist it's just empty.

Repro Method:
   1) talosctl reset all my control planes
   2) I applied the contents for inlineManifests and extraManifests to my controlplane.yaml

    inlineManifests:
      - name: argocd-namespace
        contents: |
          apiVersion: v1
          kind: Namespace
          metadata:
            name: argocd

   extraManifests:
        - https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/install.yaml

# note that i have some additional extraManifests I'm not listing
   3) talosctl apply-config on my controlplanes
   4) talosctl bootstrap on a single controlplane ip

After everything comes up:
> kubectl get pods -n argocd                                                                           
No resources found in argocd namespace.

> kubectl get pods                                                                                            
NAME                                                READY   STATUS    RESTARTS   AGE
argocd-application-controller-0                     1/1     Running   0          13m
argocd-applicationset-controller-7d4d7c7b89-q4ptx   1/1     Running   0          13m
argocd-dex-server-cbddb9676-5bwpt                   1/1     Running   0          13m
argocd-notifications-controller-7b55c64b69-k649j    1/1     Running   0          13m
argocd-redis-68bc658cfb-tml4b                       1/1     Running   0          13m
argocd-repo-server-56c67cf674-jg2wm                 1/1     Running   0          13m
argocd-server-66b7d96445-9qjc9                      1/1     Running   0          13m

> kubectl get namespace                                                                               
NAME                            STATUS   AGE
argocd                          Active   3m58s
default                         Active   4m53s
kube-node-lease                 Active   4m53s
kube-public                     Active   4m53s
kube-system                     Active   4m53s
kubelet-serving-cert-approver   Active   3m58s

Almost there but not quite.  I certainly don't see from the yaml syntax how the namespace would be associated with the extraManifest.  Maybe there's a trick to it?

Thanks in advance.

Iheanacho-ai:
fixes #647

### DIFF
--- a/public/kubernetes-guides/advanced-guides/deploy-argocd.mdx
+++ b/public/kubernetes-guides/advanced-guides/deploy-argocd.mdx
@@ -1,6 +1,6 @@
 ---
 title: "Deploy Argo CD"
-description: "In this guide you will learn how to deploy Argo CD on Talos Linux using extraManifests, or on Omni-managed clusters using manifest sync."
+description: "In this guide you will learn how to deploy Argo CD on Talos Linux with kubectl or at bootstrap time, or on Omni-managed clusters using manifest sync."
 ---
 
 import { k8s_release } from '/snippets/custom-variables.mdx';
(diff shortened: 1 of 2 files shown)


public/kubernetes-guides/advanced-guides/deploy-argocd.mdx
```diff
@@ -20,55 +20,122 @@ Before you begin, ensure you have the following:
 
 ## Installation on Talos
```
smira: I think this looks a bit strange as a heading - when using Omni one is using Talos as well, so I would put Omni first, and reword Talos as "Talos without Omni" or something close (?)

public/kubernetes-guides/advanced-guides/deploy-argocd.mdx
```diff
@@ -20,55 +20,122 @@ Before you begin, ensure you have the following:
...
+    
+    **Step 4.** Generate the patch file, embedding the manifest under `contents`:
+    
+    ```bash
+    cat <<EOF > argocd-inline-manifest.yaml
+    cluster:
+      inlineManifests:
+        - name: argocd
+          contents: |
+    $(sed 's/^\(.\)/          \1/' argocd-complete.yaml)
```
smira: `yq` should be way better way to do this ?

public/kubernetes-guides/advanced-guides/deploy-argocd.mdx
```diff
@@ -18,59 +18,121 @@ Before you begin, ensure you have the following:
...
+      --talosconfig=./talosconfig
+    ```
+
+    **Step 7.** If this is a new cluster, bootstrap it. Run this **only once** per cluster — if you have already bootstrapped, skip to the reboot command below:
+
+    ```bash
+    talosctl bootstrap --nodes <control-plane-ip> --talosconfig=./talosconfig
+    ```
+
+    If you patched an existing cluster, reboot the control plane nodes instead:
```
smira: there is no need to reboot Talos ever on such machine config changes, Talos will apply the new manifest immediately.

In fact, it doesn't matter if it was bootstrapped before or after that.

Same below

sterlingkoch (approved):

PR #670, opened 2026-07-24
docs: talos #13855 - dedicated system volumes

majabojarska:
Updates the docs for system volume configuration to cover the latest changes in regard to provisioning them on dedicated partitions, instead of having them directory-backed (under `EPHEMERAL`).

Relates to:
- https://github.com/siderolabs/talos/pull/13655
- https://github.com/siderolabs/talos/issues/10678


smira (commented):
cool, I would also add a section here: https://docs.siderolabs.com/talos/v1.13/configure-your-talos-cluster/storage-and-disk-management/disk-management/layout#disk-layout

smira (approved):
lgtm 👍

PR #675, opened 2026-07-28
docs: document the Auditor role and the audit log exporter

utkuozdemir:
Omni has a new Auditor role that grants read access plus the audit log and nothing else, so reading the audit log no longer requires Admin. Access is granted by exact role rather than by seniority, so Operator does not get it despite being more privileged, and the role cannot be assigned by an access policy or a SAML label rule.

A new page covers the audit log exporter: how to run it, where an export starts, how it resumes across restarts, and what stops it rather than being retried. The existing audit log page said Omni had no built-in way to export, so it now points at the new page instead. The event reference also drops two fields that no longer exist and gains the resource id and the audit log access event type.

The agent guide is corrected while here. It described the navigation as being edited directly in the generated file rather than in the per-product configs it is built from, and it now also lists the checks CI runs so they can be run before pushing.

Draft until the Auditor role lands in Omni and ships in a release, since the docs site publishes continuously and would otherwise describe a role that cannot be selected yet.

Part of siderolabs/omni#2985.

### DIFF
--- a/public/omni/cluster-management/export-audit-logs.mdx
+++ b/public/omni/cluster-management/export-audit-logs.mdx
@@ -1,0 +1,116 @@
+---
+title: Export Audit Logs
+description: Stream the Omni audit log to a log shipper or file with the audit log exporter.
+---
+
+The [Omni audit log exporter](https://github.com/siderolabs/omni-audit-log-exporter) is a small daemon that follows the audit log of an Omni instance and writes each event to stdout as a line of JSON. It keeps up with new events as they happen, and resumes where it left off across reconnects and restarts.
+
+Use it when audit events need to live somewhere other than Omni, such as a log aggregation system or storage that outlasts Omni's retention window.
+
+For reading the audit log interactively instead, see [Audit Logs](./using-audit-log).
+
+## Prerequisites
+
+You need:
+
+- Omni v1.10.0 or later. Following the audit log arrived in that release. Earlier versions can serve the audit log but not follow it, and the exporter exits rather than retrying.
+- The audit log enabled on that instance. On self-hosted instances it is off by default, so enable it first.
+
+## Set up the exporter
+
+Setting the exporter up takes two steps: give it an identity to authenticate with, then start it against your instance.
+
+### Step 1: Create a service account
+
+Create a service account for the exporter:
+
+```bash
+omnictl serviceaccount create --use-user-role=false --role=Auditor audit-log-exporter
+```
+
+The `Auditor` role grants read-only access to Omni's resources, plus permission to read the audit log. It grants nothing beyond that: an `Auditor` cannot create, change or delete anything, and it cannot turn audit logging on or off. Avoid giving the exporter the `Admin` role, which would also let it manage users and service accounts and grant it administrative access to every managed Kubernetes cluster.
+
+<Note>
+`--use-user-role` defaults to true, which clones the role of the user running the command instead of the role you pass. Set it to false, as above, or the account will end up with your own role.
+</Note>
+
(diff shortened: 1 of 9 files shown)


utkuozdemir:
@Iheanacho-ai Let's keep this not merged until we release Omni 1.10.0.

public/omni/cluster-management/export-audit-logs.mdx
```diff
@@ -0,0 +1,109 @@
...
+Use it when audit events need to live somewhere other than Omni, such as a log aggregation system or storage that outlasts Omni's retention window.
+
+For reading the audit log interactively instead, see [Audit Logs](./using-audit-log).
+
+## Before you begin
+
+You need:
+
+- An Omni instance with the audit log enabled. On self-hosted instances audit logging is off by default, so enable it first.
+- An Omni version whose management API supports following the audit log. Older versions can serve the audit log but not follow it, and the exporter exits rather than retrying.
```
Iheanacho-ai: what is this version

public/omni/cluster-management/export-audit-logs.mdx
```diff
@@ -0,0 +1,109 @@
...
+- A service account with the `Auditor` role. On Omni versions that predate that role, use `Admin` instead. Following the audit log arrived before the `Auditor` role did, so an instance can support the exporter while still rejecting `Auditor` as an unknown role.
```
Iheanacho-ai: since we explain how to create the auditor service account in the next step, i dont think there is any need to add this in the prerequisities section

public/omni/cluster-management/export-audit-logs.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+## Create a service account
+
+Create a service account for the exporter:
+
+```bash
+omnictl serviceaccount create --use-user-role=false --role=Auditor audit-log-exporter
+```
+
+The `Auditor` role grants read access plus the audit log, and nothing else. Avoid giving the exporter the `Admin` role, which would also let it manage users and service accounts and grant it administrative access to every managed Kubernetes cluster.
```
Iheanacho-ai: read access plus the audit log? in waht sense read access to audit log, or read access and the ability to manage the audit log ( create and read)?

Iheanacho-ai:
The hierachy of your doc is off. This is a better hierachy:

- Introduction
- Prerequisites
- Set up the exporter (parent for the actual procedure)
    - Step 1: Create a service account
    - Step 2: Run the exporter
- Configure behavior (parent for the tuning topics)
    - Choose where the export starts
    - Resume across restarts
- Understand the delivery guarantees
- Troubleshoot

public/omni/cluster-management/export-audit-logs.mdx
utkuozdemir: clarified

public/omni/cluster-management/export-audit-logs.mdx
utkuozdemir: removed

public/omni/cluster-management/export-audit-logs.mdx
utkuozdemir: made this paragraph clearer.

utkuozdemir:
> The hierachy of your doc is off. This is a better hierachy:
> 
> * Introduction
> * Prerequisites
> * Set up the exporter (parent for the actual procedure)
>   
>   * Step 1: Create a service account
>   * Step 2: Run the exporter
> * Configure behavior (parent for the tuning topics)
>   
>   * Choose where the export starts
>   * Resume across restarts
> * Understand the delivery guarantees
> * Troubleshoot

reworked as you suggested

Iheanacho-ai (approved):

Issue #651, opened 2026-07-16
How-to: Scan TEL images with SBOM and VEX


sterlingkoch:
Part of #649. **Drafted in #677** and ready to merge with it.

The core TEL scanning workflow: accurate vulnerability results using the per-schematic SBOM plus Sidero-curated VEX.

Scope:

- `grype sbom:image.spdx.json`, basic SBOM scan.
- `grype sbom:image.spdx.json --vex talos.vex.json`, scan with non-applicable CVEs filtered.
- On-demand scan reports from EIF and when to use them vs. scanning locally; the live-database caveat (results reflect request time, not build time).
- Feeding results to admission controllers and compliance tooling (brief, with links out).

**Dropped from scope, September 3:** how customer-provided extensions are attributed in the SBOM. That attribution doesn't exist. Custom extensions are reachable only through a per-deployment factory setting that no customer can configure today, so there is nothing accurate to write. The whole question is tracked on #652.

Diátaxis type: how-to.

sterlingkoch:
Customer evidence for this page: [Pylon #874](https://app.usepylon.com/issues?issueNumber=874) (July 15) — a customer followed the current SBOM scanning guide (`talos/v1.13/advanced-guides/SBOM#scanning-sboms`), hit the "VEX database is available to Enterprise customers… contact Sidero support" line, and filed a ticket asking for access. Support's answer: VEX is delivered through Enterprise Image Factory, not a separate database or a support-provisioned handoff.

Two implications for this how-to:

1. The scanning walkthrough should show the EIF retrieval path explicitly (fetch VEX from your EIF account → `grype --vex`), so the enterprise reader never needs a support ticket to complete the loop.
2. When this page lands, the existing SBOM advanced guide's "contact Sidero support" line should be updated to link here / to the EIF overview (docs#652) instead — that line is currently generating tickets that route customers through support for something that should be self-serve.

Issue #650, opened 2026-07-16
How-to: Download images and artifacts from Enterprise Image Factory


sterlingkoch:
Part of #649. **Drafted in #677.** Only the credential mechanics are still pending.

Task-oriented guide for downloading TEL artifacts from EIF (`factory.siderolabs.com`), via the web UI and programmatically. The existing EIF page explains *what* the artifacts are; this page shows *how* to get them.

Scope:

- Downloading boot assets: ISO, disk images, UKI, installer images.
- Downloading SBOM (SPDX 2.3 JSON), VEX (VEX JSON), and vulnerability scan reports (`.json`, `.table`, `.sarif`, `.cdx`), all served from the same page as the image's boot assets.
- Programmatic retrieval with `curl` for CI pipelines.
- Artifact checksums.

Out of scope: credential setup and management (#653; link to that page rather than documenting auth here).

**Status, September 3.** The page is written and correct except for how the `curl` examples pass credentials. Factory authentication is being rebuilt into a single unified API token, so the current `-u "${EIF_USERNAME}:${EIF_TOKEN}"` form will change. Rather than hold the page for it, the credential preamble moves into a snippet under `/snippets/` that every example references, so one file changes when the token lands. Discussion is on the #677 review thread.

Diátaxis type: how-to.

PR #677, opened 2026-07-29
docs: add talos enterprise linux docs

Iheanacho-ai:


public/tel/v1.0/guides/download-images-and-artifacts.mdx
```diff
@@ -0,0 +1,119 @@
...
+| Disk image (qcow2) | `metal-<arch>.qcow2` | `metal-amd64.qcow2` |
+| UKI (UEFI, SecureBoot-compatible) | `<platform>-<arch>[-secureboot]-uki.efi` | `metal-amd64-uki.efi` |
+| Kernel | `kernel-<arch>` | `kernel-amd64` |
+| Kernel command line | `cmdline-<platform>-<arch>[-secureboot]` | `cmdline-metal-amd64` |
+| Initramfs | `initramfs-<arch>.xz` | `initramfs-amd64.xz` |
+
+For example, to download a disk image with `curl`:
+
+```shell
+curl -u "${EIF_USERNAME}:${EIF_TOKEN}" -fJLO \
```
smira, line 71: I would hold on this, there is a change incoming to change this flow a bit (how auth works).

tel-v1.0-test.tmp
sterlingkoch, line 1: Pretty sure this file isn't supposed to be included.

public/tel/v1.0/guides/download-images-and-artifacts.mdx
```diff
@@ -0,0 +1,119 @@
+---
+title: "Download Images and Artifacts From Enterprise Image Factory"
+description: "How to download Talos Enterprise Linux boot assets, SBOMs, VEX data, and vulnerability scan reports from Enterprise Image Factory."
+---
+
+import { version } from '/snippets/custom-variables.mdx';
+
+Enterprise Image Factory (`factory.siderolabs.com`) serves Talos Enterprise Linux boot assets along with the artifacts unique to it: per-schematic SBOMs, VEX data, and vulnerability scan reports. This page shows how to download each of them, through the web UI and programmatically.
+
+For what these artifacts are and how Enterprise Image Factory differs from the public Image Factory, see <a href={`../../../talos/${version}/learn-more/enterprise-image-factory.mdx`}>Enterprise Image Factory</a>
```
sterlingkoch, line 10: I _think_ we need to remove the .mdx extension on five of the links on this PR.  I tried it with the .mdx and Mintlify did a 308 and redirected to the docs landing page.  I'm assuming it was an accident and not some intentional thing that I don't understand.

public/tel/v1.0/guides/download-images-and-artifacts.mdx
```diff
@@ -0,0 +1,119 @@
+---
+title: "Download Images and Artifacts From Enterprise Image Factory"
+description: "How to download Talos Enterprise Linux boot assets, SBOMs, VEX data, and vulnerability scan reports from Enterprise Image Factory."
+---
+
+import { version } from '/snippets/custom-variables.mdx';
```
sterlingkoch, line 6: .mdx (see comment below)

public/tel/v1.0/guides/download-images-and-artifacts.mdx
```diff
@@ -0,0 +1,119 @@
...
+sha256sum -c metal-amd64.raw.zst.sha256
+```
+
+The checksum verifies the integrity of the asset as served to you — that your download arrived intact. Because some images are generated dynamically and are not byte-for-byte reproducible across builds, checksums are not stable identifiers to pin over time. For cryptographic verification of provenance, use signatures instead.
+
+## See also
+
+For more information, see also:
+
+- <a href={`../../../talos/${version}/learn-more/enterprise-image-factory.mdx`}>Enterprise Image Factory</a>: To learn what these artifacts are and how the enterprise service differs from the public one
```
sterlingkoch, line 118: .mdx

public/tel/v1.0/guides/scan-tel-with-sbom-and-vex.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+Run Grype directly against the SBOM:
+
+```shell
+grype sbom:image.spdx.json
+```
+
+Grype matches every component listed in the SBOM against its vulnerability database and reports the CVEs it finds, with severity, fixed-in version, and match details.
+
+Because the SBOM is generated per schematic, it describes your exact image — the Talos base plus every system extension in your configuration. This differs from the release SBOMs published on GitHub, which cover only the unmodified base OS. For base-OS SBOMs and reading SBOMs from a running node, see <a href={`../../../talos/${version}/advanced-guides/SBOM.mdx`}>SBOMs</a>.
```
sterlingkoch, line 49: .mdx link

public/tel/v1.0/guides/scan-tel-with-sbom-and-vex.mdx
```diff
@@ -0,0 +1,79 @@
...
+- **Download a factory scan report**: Enterprise Image Factory runs the scan using its vulnerability database at the time you request the report. This requires no local tooling, and the database is always up to date. However, results can change over time as new vulnerabilities are added to the database.
+
+Both approaches produce the same type of report. Choose local scanning if you need reproducible results, or a factory scan report if you want the latest vulnerability data without any local setup.
+
+## See also
+
+For more information, see also:
+
+- [Download images and artifacts from Enterprise Image Factory](./download-images-and-artifacts): How to retrieve the SBOM, VEX data, and scan reports used on this page
+- <a href={`../../../talos/${version}/learn-more/enterprise-image-factory.mdx`}>Enterprise Image Factory</a>: What these artifacts are and how the enterprise service differs from the public one
```
sterlingkoch, line 77: .mdx link

public/tel/v1.0/guides/download-images-and-artifacts.mdx
sterlingkoch: This is the only place the auth flow touches the page, and it's the same two variables in six examples. Rather than hold the whole section for it, let's pull the credential preamble into a snippet and reference it from each example, so when unified tokens land it's one file to change and every TEL page follows.

sterlingkoch (commented):
Approving the structure. The versioned tel/v1.0 section is right and both pages read well, so this sets a good precedent for the six that follow. Line comments cover the details; three things at the PR level:

- On the conflict, both files it touches resolve the same way and neither is a hand-merge. public/docs.json is generated, so rebase on main and re-run make docs.json rather than editing it. The Makefile conflicts because the version list was reordered to newest-first after you branched, so tel-v1.0.yaml just needs re-adding after omni.yaml in each of the four targets against the current ordering.

- On the .mdx links: worth fixing carefully rather than quickly. The 308 you land on isn't a link-format quirk, it's a site-wide setting that turns every unresolvable URL into a redirect to the product landing page, filed as #723. When that gets turned off these will return honest 404s, so they need to be right.
Add Closes #650 and Closes #651 to the description so those don't stay open and unlinked after merge.

- Left a note on Andrey's comment about the auth flow. Short version is we should extract the credential lines into a snippet so this doesn't wait on the token rework.

sterlingkoch:
Superseded. The versioned public/tel/v1.0/ structure was replaced by the unversioned tree in #763, and both pages here ship at the new paths in #751 and #754. Closing rather than reworking.

The credential-preamble snippet idea from my Sept 3 comment still stands; carrying it into the deduplication issue from #759 so it isn't lost.

PR #682, opened 2026-07-29
docs: remove the multi-doc patch in the hybrid cluster doc

Iheanacho-ai:


public/omni/cluster-management/create-a-hybrid-cluster.mdx
smira, line 44: this will be a bad patch for Talos 1.13+

better to add both patch versions here: https://docs.siderolabs.com/talos/v1.13/getting-started/what's-new-in-talos#kubespan-configuration

The section below which is removed is correct - it for some reason mentions talosctl apply-config with Omni which makes no sense

smira (approved):

sterlingkoch (approved):

PR #684, opened 2026-07-29
remote

Iheanacho-ai:


sterlingkoch (commented):
This page exists on v1.11–v1.14; should we also apply the same change on v1.14 (and the others if you're touching those trees)?

Also, can we give this a better name that "Remote"?

sterlingkoch (approved):

PR #707, opened 2026-08-18
docs: add the etcd flag

Iheanacho-ai:


Iheanacho-ai:
@smira what do you think of this change

smira (commented):
lots of unrelated changes and noise for me

can we have a specific PR with a fix with a clear diff, commit message, PR description which makes sense, etc? 

There is no step or whatever that mentions something like this https://docs.siderolabs.com/omni/self-hosted/run-omni-on-prem#7-2-create-the-sqlite-storage-directory but for etcd.

For some reason SQLite home directory section is gone?

Iheanacho-ai:
> lots of unrelated changes and noise for me
> 
> can we have a specific PR with a fix with a clear diff, commit message, PR description which makes sense, etc?
> 
> There is no step or whatever that mentions something like this https://docs.siderolabs.com/omni/self-hosted/run-omni-on-prem#7-2-create-the-sqlite-storage-directory but for etcd.
> 
> For some reason SQLite home directory section is gone?

The SQLite home directory step got merge with the etcd directory step, and as thus I changed the step to  "Create storage directories"

But I just created a new PR with only that step @smira, and in that PR i decided to split the step: https://github.com/siderolabs/docs/pull/708

The rest of the changes will also carry their own PR

sterlingkoch (commented):
@Iheanacho-ai - as far as I can tell, this is redundant with #708, #709, and #710.  We should close this one and then merge those in this order: #708 -> #710 -> #709.

PR #708, opened 2026-08-18
docs: mount etcd data directory in self-hosted Omni guide

Iheanacho-ai:
## What

Adds the missing etcd data volume mount to the embedded-etcd docker run command in the self-hosted Omni deployment guide.

## Why

Without this mount, etcd data exists only inside the container's writable layer. Recreating the container, e.g., during an upgrade, wipes all Omni cluster state without warning. Flagged by @smira, confirmed by @artem.chernyshev.

## Change

- New Step 7.3, "Create the etcd storage directory," structured identically to the existing SQLite step (7.2)
- One added `-v` mount in the Embedded etcd docker run command (7.5)
- Steps 7.4 (EULA) and 7.5 (Start Omni) are renumbered from 7.3/7.4 as a result, content unchanged, only the numbers and two in-step cross-references shifted
- External etcd tab is untouched, it already disables embedded etcd and never needed this mount

## Testing

Verified against a live EC2 instance: recreated the `omni` container after applying this fix and confirmed via `docker logs` (existing user preserved, not re-initialized) and the etcd data directory (unchanged contents/structure) that cluster state survived the recreate.

### DIFF
--- a/public/omni/self-hosted/run-omni-on-prem.mdx
+++ b/public/omni/self-hosted/run-omni-on-prem.mdx
@@ -344,46 +344,55 @@ The commands below use CLI flags to start Omni. If you prefer to use a configura
 <Tabs>
   <Tab title="Embedded etcd (default)">
 
-```bash
-docker run -d \
-  --name omni \
-  --net=host \
-  --cap-add=NET_ADMIN \
-  --device /dev/net/tun:/dev/net/tun \
-  --restart=unless-stopped \
-  -v $(pwd)/ca.pem:/etc/ssl/certs/ca-certificates.crt:ro,Z \
-  -v $(pwd)/server-key.pem:/server-key.pem:ro,Z \
-  -v $(pwd)/server-chain.pem:/server-chain.pem:ro,Z \
-  -v $(pwd)/omni.asc:/omni.asc:ro,Z \
-  -v $HOME/sqlite:/_out/sqlite:rw,Z \
-  ghcr.io/siderolabs/omni:${OMNI_VERSION} \
-    --name=omni \
-    --cert=/server-chain.pem \
-    --key=/server-key.pem \
-    --machine-api-cert=/server-chain.pem \
-    --machine-api-key=/server-key.pem \
-    --machine-api-bind-addr=0.0.0.0:8090 \
-    --private-key-source=file:///omni.asc \
-    --event-sink-port=8091 \
-    --bind-addr=0.0.0.0:443 \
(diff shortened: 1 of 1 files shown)


public/omni/self-hosted/run-omni-on-prem.mdx
```diff
@@ -287,7 +287,15 @@ This command creates the directory on the host that will be mounted into the con
 mkdir -p $HOME/sqlite
 ```
 
-### 7.3: Configure EULA acceptance
+### 7.3: Create the etcd storage directory
+
+Omni uses etcd for cluster state. This command creates the directory on the host that will be mounted into the container:
```
smira: only for embedded etcd? (idk if it should be in the docs)

the SQLile once above say `$HOME`, this one says `$(pwd)` (that's a totally different syntax? why?)

public/omni/self-hosted/run-omni-on-prem.mdx
```diff
@@ -355,6 +363,7 @@ docker run -d \
   -v $(pwd)/server-key.pem:/server-key.pem:ro,Z \
   -v $(pwd)/server-chain.pem:/server-chain.pem:ro,Z \
   -v $(pwd)/omni.asc:/omni.asc:ro,Z \
+  -v $(pwd)/etcd:/_out/etcd:rw,Z \
```
smira: why `$(pwd)` vs. `$HOME` below (idk?)

Iheanacho-ai: just changed it to HOME, this too @smira 

https://github.com/siderolabs/docs/pull/709 for the ca trust

public/omni/self-hosted/run-omni-on-prem.mdx
```diff
@@ -300,7 +300,7 @@ If your organization has a separate agreement with Sidero Labs, that agreement t
 <Tabs>
   <Tab title="CLI flags">
 
-  You can pass your EULA details as CLI flags when starting your Omni instance in **Step 7.4**:
+  You can pass your EULA details as CLI flags when starting your Omni instance in **Step 7.5**:
```
sterlingkoch: One leftover before I approve: the EULA section's CLI-flags tab says "Step 7.5" but Start Omni is still 7.4 (the browser tab below correctly says 7.4). Looks like a holdover from when the etcd dir was its own numbered step. Revert that to 7.4 and this is good to go.

sterlingkoch (changes requested):

sterlingkoch (approved):

PR #711, opened 2026-08-18
docs: warn Docker users to verify etcd mount before upgrading Omni

sterlingkoch:
Adds a warning to the Upgrade Omni page for Docker-based deployments created while the Run Omni On-Prem guide was missing the etcd volume mount (introduced in #448, fixed in #708).

The upgrade page is the page an affected user is most likely following when they hit the destructive action — recreating the container destroys all Omni state if etcd data was never mounted from the host. The warning inlines a one-command `docker inspect` check so the reader can verify on the spot, and directs affected users to the status page and support rather than attempting self-service recovery.

A dedicated recovery page (identification + step-by-step data migration) is in progress separately; when it lands, the warning's status-page reference will be repointed at it.

### DIFF
--- a/public/omni/self-hosted/upgrading-omni.mdx
+++ b/public/omni/self-hosted/upgrading-omni.mdx
@@ -5,19 +5,12 @@ description: Prepare, validate, and apply a new Omni release without risking dat
 
 This guide explains the supported process for upgrading an on-prem Omni installation.
 
-<Warning>
+<Danger>
 **Before upgrading a Docker-based deployment:** if you installed Omni with `docker run`
-between April and August 2026, check whether your etcd data is mounted from the host:
-
-```bash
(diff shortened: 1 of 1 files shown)


public/omni/self-hosted/upgrading-omni.mdx
```diff
@@ -5,6 +5,20 @@ description: Prepare, validate, and apply a new Omni release without risking dat
 
 This guide explains the supported process for upgrading an on-prem Omni installation.
 
+<Warning>
```
Iheanacho-ai: i think this would be better as a Danger call out https://www.mintlify.com/docs/components/callouts

also link to the recovery doc

Iheanacho-ai (approved):

PR #714, opened 2026-08-20
feat: add doc-accuracy reviewer to catch harmful documentation changes

Iheanacho-ai:
## What

Adds `make check-doc-accuracy`, a reviewer that checks docs for changes that could harm a reader who follows them (commands that lose data, destroy things, or weaken security, not just typos). Also runs it in CI on every docs PR, forks included, as an advisory check.

## Why

The worst doc bugs are commands that run fine but still do damage, like a service started without its data mount or a dropped safeguard. They pass every syntax check, so a flag checker won't catch them. You need something that reads the snippet and reasons about what happens if you run it. Until now that only ran when someone remembered to run it locally.

## Change

- `tools/doc-accuracy`: a Go program that drives the `claude` CLI as a read-only reviewer (it reports, never edits). It reviews the `.mdx` files you changed and checks claims about Talos/Omni against the upstream source.
- Looks for harm first: data loss, destructive commands, removed safeguards, security downgrades, then plain wrong flags and values.
- Output: a short terminal summary, a JSON file (verdict + a reason and fix per finding), and in CI, inline annotations plus one summary comment that updates in place on each push.
- CI (`.github/workflows/doc-accuracy.yaml`): uses `pull_request_target` so it works on fork PRs. Checks out our base branch only, never the fork's code, and reads the PR as a text diff. Nothing from the PR is run. It never blocks merge.

## Usage

```bash
make check-doc-accuracy                                # files you changed vs HEAD
make check-doc-accuracy DOC=public/omni/....mdx        # one specific file
make check-doc-accuracy DOC_ACCURACY_BASE=origin/main  # PR-style, vs a branch
make check-doc-accuracy-all                            # the whole public/ tree (slow)
```

## Testing

- `go test`, `go vet`, and `gofmt` all pass.
- Ran it against a sample doc with four planted problems (etcd with no volume mount, rm -rf /, --insecure on a bootstrapped node, a stale image tag). It caught all four at the right severity and checked the current Talos version upstream. Verdict FAIL, exit 1.
- Confirmed only a CRITICAL fails the run; warnings and notices don't. No changed docs pass cleanly, and a missing claude CLI gives a clear message.
- Not yet tested on a live PR, since pull_request_target runs the workflow from main, so the annotations and comment only show up once this merges. The `ANTHROPIC_API_KEY` secret is set.

### DIFF
--- a/.github/workflows/doc-accuracy.yaml
+++ b/.github/workflows/doc-accuracy.yaml
@@ -1,0 +1,262 @@
+name: Doc Accuracy
+
+# AI accuracy/harm review of changed docs. Advisory: it annotates the diff and
+# posts a summary comment for findings, but does not block the PR.
+#
+# This runs on PRs from forks too, which is why it uses `pull_request_target`
+# rather than `pull_request`: only `pull_request_target` exposes repo secrets
+# (the Anthropic key) and a write token to a fork PR. That trigger is safe *only*
+# because of how this job is built — see the security notes on the checkout and
+# diff steps below. The short version: the fork's ref is never checked out and
+# its content is never executed; the PR's changes reach the reviewer only as
+# text, to reason about, exactly like pasting a doc into a chat and asking "does
+# this look right?".
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    paths:
+      # Only doc changes. Deliberately *not* tools/doc-accuracy/** — under
+      # pull_request_target the job runs the base branch's copy of the tool, so a
+      # PR to the tool itself would run the old tool against zero changed docs.
+      # Test tool changes with `make check-doc-accuracy` locally.
+      - "public/**/*.mdx"
+
+# Superseded runs on the same PR are cancelled so only the latest review stands.
+concurrency:
+  group: doc-accuracy-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  doc-accuracy:
+    runs-on: ubuntu-latest
+    # Each batch of files is a separate model run with its own 10-minute budget,
+    # so the default 6-hour job limit is no real bound. This one is: past it the
+    # run is killed and reported as INCOMPLETE by the comment step below.
+    timeout-minutes: 30
+    steps:
+      - name: Checkout the BASE branch (never the PR's fork ref)
+        uses: actions/checkout@v4
+        with:
+          # SECURITY: actions/checkout persists its token in .git/config as an
+          # extraheader by default. This job then points a Read/Grep-capable
+          # model at that very directory while it chews on attacker-controlled
+          # diff text, so don't leave a second credential lying inside it —
+          # nothing here needs git authentication after the clone.
+          persist-credentials: false
+        # SECURITY: do NOT add `ref: ${{ github.event.pull_request.head.sha }}`
+        # (or any fork ref) here. Under pull_request_target this job has repo
+        # secrets and a write token; checking out the fork would put untrusted
+        # code on a runner that holds them — the classic pull_request_target
+        # exploit. The default checkout is the base branch, which is exactly what
+        # we want: it is the trusted source-of-truth the reviewer reads to ground
+        # its checks. The PR's own changes come in as text in the next step.
+
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          # Must match tools/doc-accuracy/go.mod (newer than the other docs jobs).
+          go-version: "1.25.1"
+
+      - name: Install Claude Code CLI
+        # Pinned deliberately: the reviewer's read-only + egress-scoped guarantee
+        # rests on the CLI honouring `--tools` and `--allowedTools` (WebFetch is
+        # domain-scoped to GitHub there). A silent upgrade that changed that
+        # behaviour must not slip in unnoticed — bump this after re-checking.
+        run: npm install -g @anthropic-ai/claude-code@2.1.226
+
+      - name: Fetch the PR changes as a text diff (no fork checkout)
+        env:
+          GH_TOKEN: ${{ github.token }}
+          # A PR number is an integer and a head SHA is hex — safe to interpolate.
+          # PR *title/body/branch* are attacker-controlled and are never placed
+          # on a shell line; nothing here reads them.
+          PR: ${{ github.event.pull_request.number }}
+        run: |
+          set -euo pipefail
+          # `gh pr diff` fetches the diff through the API as plain text. It does
+          # NOT check out or fetch the fork's commits into the working tree — the
+          # fork's content only ever exists here as this diff file (data), which
+          # the reviewer reads and never executes.
+          gh pr diff "$PR" --repo "$GITHUB_REPOSITORY" > "$RUNNER_TEMP/pr.diff"
+          echo "Fetched PR diff: $(wc -l < "$RUNNER_TEMP/pr.diff") lines."
+
+      - name: Review the changed docs for accuracy (advisory)
+        id: review
+        continue-on-error: true # advisory: a FAIL annotates/comments but never blocks
+        env:
+          # GUARDRAIL: keep this to the single dedicated key. The reviewer's
+          # WebFetch is domain-scoped to GitHub (see tools/doc-accuracy), so it
+          # cannot exfiltrate this env to an arbitrary host — but don't lean on
+          # that to add other secrets here (a write-scoped token especially).
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+        # -diff-file feeds the changed docs from the fetched diff (as data); the
+        # reviewer is told not to read them from disk (the working tree is the
+        # base, not the change) but may Read/Grep the base checkout and WebFetch
+        # the upstream Talos/Omni GitHub source to ground its findings (that
+        # WebFetch is domain-scoped to GitHub in the tool). -format github emits
+        # inline ::error/::warning annotations. -fetch=false: nothing to refresh.
+        # -max-files caps the fan-out: a PR touching more docs than that is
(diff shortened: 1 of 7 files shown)


tools/doc-accuracy/main.go
```diff
@@ -0,0 +1,405 @@
...
+// review regardless of how the model split it across messages.
+func runClaude(ws, model, prompt string, out io.Writer) error {
+	args := []string{"-p"}
+	if model != "" {
+		args = append(args, "--model", model)
+	}
+	args = append(args,
+		"--output-format", "stream-json", "--verbose",
+		"--permission-mode", "bypassPermissions",
+		"--allowedTools", "Read", "Grep", "Glob", "WebFetch",
```
sterlingkoch: From Claude:
`--allowedTools` is ignored under `--permission-mode bypassPermissions` (claude-code#12232). The whitelist here isn't actually a whitelist, so Bash is unrestricted and the read-only claim doesn't hold. `--disallowedTools Edit,Write` should still block those two. `--permission-mode dontAsk` with this allowedTools list, or `--tools` as a real whitelist, would do what this is aiming for.

tools/doc-accuracy/reviewer-prompt.md
```diff
@@ -0,0 +1,143 @@
...
+     `wipefs`, `docker system prune`, `docker volume rm`, `kubectl delete`,
+     `DROP`/`DELETE`, `git push --force`, `truncate`, disk-wiping or
+     factory-reset flows — especially when aimed at a path/resource a reader
+     might have real data in, or shown without a clear warning.
+   - **Removed safeguard (diff-specific)** — a change that deletes a flag, line,
+     mount, `--dry-run`, confirmation prompt, backup step, or `|| exit` guard
+     that previously made the procedure safe. Compare against the diff and the
+     upstream source; a deletion that makes a command *more* dangerous is a
+     top-priority finding even if what remains is valid.
+   - **Security downgrades / exposure** — disabling TLS or auth, `--insecure`,
```
sterlingkoch, line 67: `--insecure` as CRITICAL will fail half the Talos bootstrap corpus (`talosctl apply-config --insecure`). Worth carving out the documented bootstrap case so this doesn't train us to ignore the tool.

Makefile
```diff
@@ -512,3 +512,36 @@ style-check-changed-auto: ## Check changed .mdx files, preferring local Go and f
...
+# judgment-based review you run before push, not a CI gate; it still exits
+# non-zero on critical findings so you *can* gate on it.
+#
+# DOC_ACCURACY_BASE is the ref "changed" mode diffs against (default HEAD);
+# DOC_ACCURACY_MODEL optionally overrides the model.
+DOC_ACCURACY_BASE ?= HEAD
+DOC_ACCURACY_MODEL ?=
+
+.PHONY: check-doc-accuracy
+check-doc-accuracy: ## AI-review docs for accuracy/harm. Scope one file with DOC=public/path; base with DOC_ACCURACY_BASE
```
sterlingkoch: The default base is HEAD, so after a local commit this won't actually review anything (`DOC_ACCURACY_BASE=origin/main` is in the README but easy to miss). Defaulting to origin/main (with a fallback) would match how this will actually get run.

.github/workflows/doc-accuracy.yaml
```diff
@@ -0,0 +1,173 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          PR: ${{ github.event.pull_request.number }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
+          MARKER='<!-- doc-accuracy-comment -->'
+          FINDINGS="_out/doc-accuracy-findings.json"
+          # If the review step never produced a file (e.g. it errored early),
```
sterlingkoch: If the review step errors (timeout, claude crash, no verdict), this manufactures a PASS. Combined with the "existing comment" branch below, a failed run will overwrite a previous FAIL with "Nothing flagged."

Gate this on steps.review.outcome. If the review step didn't succeed, post "reviewer didn't finish, see the log" and leave the last real comment alone. A missing findings file is not a clean review.

Iheanacho-ai: Good catch, fixed. The tool now emits a tri-state verdict (PASS/FAIL/INCOMPLETE); a crash or no-verdict run is INCOMPLETE and posts a separate 'didn't finish' note instead of overwriting the findings comment

sterlingkoch:
Review pass. Leading with the good news, because I went in expecting the classic `pull_request_target` hole and it isn't there.

**A fork contributor cannot execute code on the privileged runner.** The checkout at [`.github/workflows/doc-accuracy.yaml:34-42`](https://github.com/siderolabs/docs/blob/9ed27d3b67152c2fd615522d48e2630863af524e/.github/workflows/doc-accuracy.yaml#L34-L42) has no `ref:`, so the tree is the base branch and the fork's ref is never fetched. Fork content enters only as text via `gh pr diff`, and I traced every path from that data to `exec.Command` and found none. In `dataOnly` mode `main.go:206-214` stitches diff sections rather than taking the `git diff -- <files>` branch, so attacker-derived filenames never reach a subprocess. Diff-header forgery is blocked too, since `splitDiffByFile` only reads a `+++` before the first `@@` and hunk body lines always carry a leading space, `+`, or `-`. The session runs `--tools Read Grep Glob WebFetch` with no Bash, Write, or Edit, and WebFetch allowlisted to three GitHub hosts. `GITHUB_TOKEN` isn't in the review step's env. That's a careful design for a trigger that usually isn't.

The caveat is that prompt injection through the diff can still make the model read the key, which the code honestly admits at `main.go:1009-1017`, and the only thing stopping egress is the WebFetch domain allowlist enforced by a third-party CLI. One control carrying the whole design is why item 1 matters.

**1. Don't leave a second credential next to the first.** [`.github/workflows/doc-accuracy.yaml:34-35`](https://github.com/siderolabs/docs/blob/9ed27d3b67152c2fd615522d48e2630863af524e/.github/workflows/doc-accuracy.yaml#L34-L35). `actions/checkout` persists credentials by default, so the job's `pull-requests: write` token lands in `.git/config` as an extraheader, inside the exact directory we then point a Read/Grep-capable model at while it chews on attacker-controlled diff text. Add `with: persist-credentials: false`.

**2. No cost bound and no job timeout.** [`.github/workflows/doc-accuracy.yaml:31-32`](https://github.com/siderolabs/docs/blob/9ed27d3b67152c2fd615522d48e2630863af524e/.github/workflows/doc-accuracy.yaml#L31-L32) and [`tools/doc-accuracy/main.go:40`](https://github.com/siderolabs/docs/blob/9ed27d3b67152c2fd615522d48e2630863af524e/tools/doc-accuracy/main.go#L40). At `reviewBatchSize = 20`, #735 (157 mdx) and #733 (162 mdx) would each fan out to 8 or 9 separate Claude sessions per push, each with its own 10-minute budget, against the 6-hour default job limit. #735 is pure frontmatter backfill, which is the worst value-per-token case there is. Want a `timeout-minutes: 30` and a max-files bail-out that posts "too large to review" instead of fanning out.

**3. I don't think the inline annotations render.** [`.github/workflows/doc-accuracy.yaml:197`](https://github.com/siderolabs/docs/blob/9ed27d3b67152c2fd615522d48e2630863af524e/.github/workflows/doc-accuracy.yaml#L197). Under `pull_request_target` the check run's head SHA is the base commit, not the PR head, so annotations shouldn't attach to the Files changed tab. If that's right, this line points readers at nothing and the `details`/`fix` text (the useful half) is only visible in the run log. The whole "terse comment points at the detailed annotation" split depends on it, so worth confirming on a live run.

**4. Silent truncation.** [`tools/doc-accuracy/main.go:221`](https://github.com/siderolabs/docs/blob/9ed27d3b67152c2fd615522d48e2630863af524e/tools/doc-accuracy/main.go#L221). `truncateLines` drops everything past 4000 diff lines, and in diff-file mode the model is explicitly told not to read the changed files from disk (`main.go:831`), so it can't recover the rest and has no signal it happened. Append a "diff truncated, N lines omitted" marker so the review doesn't quietly cover less than it claims.

**5. Expect false positives until the bar is calibrated.** `tools/doc-accuracy/reviewer-prompt.md:138-144` puts "a wrong value" and "would break on copy-paste" in CRITICAL, and only CRITICAL produces the red FAIL comment. Advisory so it won't block, but plan on a 🔴 landing on a fair number of routine PRs at first.

**Minor.** `doc-accuracy.yaml:19` includes `tools/doc-accuracy/**` in `paths`, so a PR to the tool triggers the job, which under `pull_request_target` runs the base version against zero mdx files. And `reviewer-prompt.md:121-124` claims "only the text of your last message is shown," which isn't true, since `runClaude` captures every assistant text block at `main.go:1080-1086`.

**One thing to flag: this workflow has never actually run.** `pull_request_target` executes from the default branch and the file isn't on `main` yet, so the two green checks here are unrelated jobs. Annotation rendering, the comment upsert finding its prior comment, and secret availability on fork runs are all unexercised.

**Good work.** The annotation escaping is right, with `escapeData`/`escapeProp` at `main.go:573-587` turning newlines into `%0A` so a model-authored message can't forge a second `::error` command. The INCOMPLETE handling is the best part: `overallVerdict` at `main.go:419-428` returns PASS only on a completed clean run, and the workflow keys off `steps.review.outcome` rather than `conclusion` at `:103`, which is the correct choice under `continue-on-error` and easy to get wrong. A crash can't overwrite a real FAIL. The prompt has an explicit untrusted-content paragraph, tells the model to check upstream source rather than reason from memory, and carves out the `talosctl --insecure` maintenance-mode exception, which is real Talos nuance an off-the-shelf prompt would get wrong. 32 tests including path traversal and the `+++`-inside-a-hunk case. `gofmt`, `go vet`, `go build`, `go test` all clean, and the CLI pin at `@2.1.226` is the right call.

sterlingkoch (approved):

PR #730, opened 2026-09-02
docs: document SELinux labels for CSI volumes

Demivan:
## What

Documents how to give CSI-provisioned volumes an SELinux label, as a new `### CSI volumes`
subsection on the Talos SELinux page.

## Why

Talos does not mount CSI volumes, so it cannot label them and pods cannot access their contents.
The label can be set through StorageClass `mountOptions` instead, which reaches the driver as a
mount option. That was not documented anywhere.

Requested by @dsseng in
https://github.com/siderolabs/talos/issues/14182#issuecomment-5493059911 (the mount permission
this relies on landed in siderolabs/talos#14207).

NFS and SMB are deliberately not covered here, since siderolabs/talos#13987 is still in flight for
those.

## Change

Adds the subsection to `security/selinux.mdx` in v1.12, v1.13 and v1.14, per the current,
previous and upcoming versioning policy. Identical text in all three.

Also marks the `SecurityState` example on the same page as `shellsession`.
`style-check-changed` reads the whole of any changed file and rejects a `$` prompt in a `shell`
block, and that block contains command output, so it keeps the prompt.

No nav or `docs.json` change: the page is already listed.

## Testing

Ran the four `docs-ci` checks locally against `origin/main` and all four pass, with 0 errors and
0 warnings on the three changed files and `docs.json` regenerating unchanged. The two external
links were checked by hand, since `make broken-links` only resolves internal ones. Drafted with
Claude Code. The claim about which drivers pass mount options through comes from my own clusters
(OpenEBS LocalPV-LVM on ext4, OpenEBS Mayastor on xfs) rather than from upstream documentation.

### DIFF
--- a/public/talos/v1.12/security/selinux.mdx
+++ b/public/talos/v1.12/security/selinux.mdx
@@ -39,7 +39,7 @@ customization:
 
 You can query the SELinux state with:
 
-```shell
+```shellsession
 $ talosctl --nodes <IP> get SecurityState
 NODE         NAMESPACE   TYPE            ID              VERSION   SECUREBOOT   UKISIGNINGKEYFINGERPRINT   PCRSIGNINGKEYFINGERPRINT   SELINUXSTATE
 172.20.0.2   runtime     SecurityState   securitystate   1         false                                                              enabled, permissive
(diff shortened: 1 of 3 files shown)


public/talos/v1.12/security/selinux.mdx
```diff
@@ -55,6 +55,30 @@ As for version 1.10, SELinux runs in permissive mode by default, which does not
 SELinux can be put in enforcing mode (to actually prevent access when it is not authorized by the policy) by adding `enforcing=1` to the kernel cmdline.
 This is most commonly done via the configuration in the Image Factory.
 
+### CSI volumes
```
smira, line 58: v1.12 doesn't fully support it

public/talos/v1.13/security/selinux.mdx
```diff
@@ -55,6 +55,30 @@ As for version 1.10, SELinux runs in permissive mode by default, which does not
 SELinux can be put in enforcing mode (to actually prevent access when it is not authorized by the policy) by adding `enforcing=1` to the kernel cmdline.
 This is most commonly done via the configuration in the Image Factory.
 
+### CSI volumes
```
smira, line 58: this should work properly in 1.13.10+

smira (approved):

public/talos/v1.12/security/selinux.mdx
Demivan: Should I remove this from 1.12 docs?

smira: we will not backport that far, 1.12 is going out of support

PR #736, opened 2026-09-03
docs: fix the system extension link in the storage file

Iheanacho-ai:
## What

fix the system extension link in the storage file

## Why
pointing to version 1.10

## Change

made it always point to the latest file

## Testing
ran `make broken-links` and `make preview`

### DIFF
--- a/public/kubernetes-guides/cni/multus.mdx
+++ b/public/kubernetes-guides/cni/multus.mdx
@@ -3,6 +3,8 @@ title: "Deploy Multus CNI"
 description: "Learn how to install and configure Multus CNI on Talos Linux to attach multiple network interfaces to a pod."
 ---
 
+import { version } from '/snippets/custom-variables.mdx';
+
 [Multus CNI](https://github.com/k8snetworkplumbingwg/multus-cni) is a container network interface (CNI) plugin for Kubernetes that enables attaching multiple network interfaces to pods.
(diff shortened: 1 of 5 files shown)


sterlingkoch (approved):
Nice fix. There are seven more hardcoded v1.10 links in the same tree if you want to sweep them in this PR: monitoring-and-observability/etcd-metrics.mdx:37,57, advanced-guides/upgrading-kubernetes.mdx:12,14,147, csi/local-storage.mdx:68,71.

sterlingkoch:
Good fix, and the right pattern. Verified the import matches the twelve other pages using `import { version } from '/snippets/custom-variables.mdx';`, the relative depth resolves correctly (from `/kubernetes-guides/csi/storage`, `../../talos/...` lands on `/talos/...`), and the target page exists in both v1.13 and v1.14. Worth knowing that `make broken-links` can't see inside a JSX template-literal href, so its passing isn't evidence here; I checked the target by hand.

One thing while you're in this neighborhood: there are seven more hardcoded `v1.10` links in the same tree, same one-line fix.

- `public/kubernetes-guides/monitoring-and-observability/etcd-metrics.mdx:37` and `:57`
- `public/kubernetes-guides/advanced-guides/upgrading-kubernetes.mdx:12`, `:14`, `:147`
- `public/kubernetes-guides/csi/local-storage.mdx:68` and `:71`

Nit: the import is at line 8 here, and the convention in the other pages is line 6, straight after the frontmatter.

PR #738, opened 2026-09-03
docs: document the x86-64-v2 CPU requirement

### LINKED ISSUE #725: Document the x86-64-v2 CPU requirement
## Problem

Talos requires the x86-64-v2 microarchitecture level and enforces it at boot. A machine below that level halts:

```
[talos] [initramfs] x86_64 microarchitecture level: 1
[talos] [initramfs] it might be that the VM is configured with an older CPU model, please check the VM configuration
[talos] [initramfs] x86_64 microarchitecture level 2 or higher is required, halting
```

That requirement is not documented anywhere in the current docs.

- [System Requirements](https://docs.siderolabs.com/talos/latest/getting-started/system-requirements) lists memory, cores, and disk. It never mentions architecture or instruction sets, at any version from v1.10 through v1.14.
- Corpus-wide, `x86-64-v2` appears only on the Proxmox pages, and only for v1.6 through v1.10. From v1.11 on, the Proxmox page keeps the `kvm64` workaround and drops the sentence explaining what the workaround is for. A reader gets the fix without the reason.

A reader whose machine halts has no page to land on and no term to search for, which is the worst shape a documentation gap can take. It also gets misread as a missing capability rather than a stated constraint: "support for older Intel processors" turns up in user feedback as a feature request.

## What to do

1. Add the requirement to System Requirements, including the boot error text, so that someone pasting the error into search finds it. Applies to current live versions.
2. Restore the explanation on the Proxmox page next to the existing `kvm64` workaround and CPU-type table row, so the fix has its reason back.
3. Link the `GOAMD64` build option from [Building Custom Talos Images](https://docs.siderolabs.com/talos/latest/build-and-extend-talos/custom-images-and-development/building-images) to the requirement, and say what it is for. That option is documented today with no connection to the constraint it works around, so nobody who needs it knows it exists.

## Background

The boot-time check arrived in [siderolabs/talos#8537](https://github.com/siderolabs/talos/pull/8537), which introduced `constants.MinimumGOAMD64Level`. The requirement itself was announced in the [Talos v1.1.0 release notes](https://github.com/siderolabs/talos/releases/tag/v1.1.0):

> Talos is built for x86-64 architecture with support for x86-64-v2 microarchitecture level, so Talos no longer runs on processors supporting only baseline `x86-64` microarchitecture (before 2009).

Diátaxis type: reference (System Requirements), plus a how-to correction on the Proxmox page.

Iheanacho-ai:
## What
Documents Talos's x86-64-v2 CPU microarchitecture requirement across all 8 live versions (v1.7–v1.14): added to System Requirements, restored on the Proxmox `kvm64` workaround (v1.11–v1.14, where it had been dropped), and linked from the `GOAMD64` build option in Building Custom Talos Images.

## Why
Closes #725. Talos halts at boot below x86-64-v2 with no matching page to land on and no explanation for the existing `kvm64`/`GOAMD64` workarounds — readers hit the wall with no way to search their way out.

## Change
- System Requirements: new "CPU architecture (requirements)" section with the verbatim boot error text, linking to the Proxmox fix and the `GOAMD64` option.
- Proxmox (v1.11–v1.14): restored the "why" next to the `kvm64` table row and workaround.
- Building Custom Talos Images: `GOAMD64=v1` now links back to the requirement it works around.

## Testing

- Ran: `make broken-links`, `make validate-docs-nav`, `make style-check-changed STYLE_CHECK_BASE=upstream/main`, `make docs.json && git diff --exit-code`
- Where: local checkout
- By: agent (Claude)
- Result: all four pass — 0 new style errors/warnings on touched lines, no broken links, nav and docs.json unchanged/current
- Filled from outside the page: none

### DIFF
--- a/public/talos/v1.10/getting-started/system-requirements.mdx
+++ b/public/talos/v1.10/getting-started/system-requirements.mdx
@@ -72,3 +72,15 @@ Talos Linux itself only requires less than 100 MB of disk space, but the EPHEMER
 Talos manages disk partitioning automatically during installation, creating EFI, META, STATE, and EPHEMERAL partitions. The EPHEMERAL partition then expands to fill all the space left after the first three. That space can either remain entirely with EPHEMERAL or be divided into additional user volumes, depending on your needs. See [Disk Layout](../configure-your-talos-cluster/storage-and-disk-management/disk-management#disk-layout) for details.
 
 For production, it is often more efficient to dedicate a smaller disk for the Talos installation itself, and use additional disks for workload storage. Using a large, single disk for both system and workloads is supported, but may not be optimal depending on your environment.
+
+## CPU requirements
+
+Talos images for amd64 require a CPU that supports the x86-64-v2 microarchitecture level. A CPU below that level halts at boot with:
+
+```text
+[talos] [initramfs] x86_64 microarchitecture level: 1
+[talos] [initramfs] it might be that the VM is configured with an older CPU model, please check the VM configuration
+[talos] [initramfs] x86_64 microarchitecture level 2 or higher is required, halting
+```
+
+This is most often seen on virtual machines whose CPU type defaults to an older or generic model, such as Proxmox's `kvm64` — see the [Proxmox](../platform-specific-installations/virtualized-platforms/proxmox) guide for the fix. To build Talos images for CPUs that do not meet this requirement, use the `GOAMD64=v1` build option described in [Building Custom Talos Images](../build-and-extend-talos/custom-images-and-development/building-images#customizations).
--- a/public/talos/v1.10/build-and-extend-talos/custom-images-and-development/building-images.mdx
+++ b/public/talos/v1.10/build-and-extend-talos/custom-images-and-development/building-images.mdx
@@ -50,7 +50,7 @@ be overridden with:
 ## Customizations
 
 Some of the build parameters can be customized by passing environment variables to `make`, e.g. `GOAMD64=v1` can be used to build
-Talos images compatible with old AMD64 CPUs:
+Talos images for CPUs that don't meet the default [x86-64-v2 requirement](../../getting-started/system-requirements#cpu-requirements):
 
 ```bash
 make <target> GOAMD64=v1
--- a/public/talos/v1.11/build-and-extend-talos/custom-images-and-development/building-images.mdx
+++ b/public/talos/v1.11/build-and-extend-talos/custom-images-and-development/building-images.mdx
@@ -50,7 +50,7 @@ be overridden with:
 ## Customizations
 
 Some of the build parameters can be customized by passing environment variables to `make`, e.g. `GOAMD64=v1` can be used to build
-Talos images compatible with old AMD64 CPUs:
+Talos images for CPUs that don't meet the default [x86-64-v2 requirement](../../getting-started/system-requirements#cpu-requirements):
 
 ```bash
 make <target> GOAMD64=v1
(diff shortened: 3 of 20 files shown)


sterlingkoch:
Read this while it's still in draft, so just notes rather than a review. The substance holds up well.

**The version claim checks out on every page you touched.** I verified against upstream rather than the release notes alone: `GOAMD64 ?= v2` first appears in the v1.1.0 Makefile (v1.0.0 has no `GOAMD64` line at all) and is still `v2` at every tag from v1.7.0 through v1.14.0. `MinimumGOAMD64Level = 2` confirms the level is v2 and not v3, and in v1.14 that constant moved to build-tagged files (`constants/amd64.go` at level 2 by default, `constants/amd64_v1.go` at level 1 under `//go:build amd64.v1`), which independently confirms `GOAMD64=v1` really is the documented escape hatch.

**The boot error text is byte-identical to the source, and the v1.7 floor is exactly right.** The three strings live in `internal/app/init/main.go` in `cpuInfo()`, identical in v1.7.0 and v1.14.0, and the `[talos] [initramfs]` prefix comes from `kmsg.SetupLogger`. The check itself arrived in siderolabs/talos#8537, milestone v1.7, and is absent from v1.6.0. So starting the quoted error at v1.7 is correct.

**All 28 added links resolve.** I checked each one mechanically: target file exists in the same version tree, and for anchored links the heading exists in that file and slugifies to the anchor. Zero failures, no cross-version leakage. The two anchor spellings correctly track the two heading spellings, which is the seven-of-eight bug I was looking for and didn't find.

**The heading is fine.** `## CPU architecture` on v1.7 to v1.11 and `## CPU architecture requirements` on v1.12 to v1.14 correctly tracks the sentence-case rewrite those newer pages already had (`Minimum requirements`, `Storage requirements`). No existing heading was renamed, so no anchor broke.

Two nits for when you mark it ready:

1. Em dash in the new prose at `public/talos/v1.14/getting-started/system-requirements.mdx:90`, and the equivalent line in the other seven (v1.7 `:84`, v1.8 to v1.11 `:86`, v1.12 and v1.13 `:90`). House style is a period instead.
2. `don't` at `building-images.mdx:53` against `do not` in the System Requirements paragraph. Pick one.

Also worth noting the `kvm64` explanation is technically right rather than merely plausible: the flag list already on the page (`+cx16,+lahf_lm,+popcnt,+sse3,+ssse3,+sse4.1,+sse4.2`) is the x86-64-v2 feature set, so "predates the level Talos requires, add these flags" is the correct causal story.

One thing that is out of scope here, filing separately: `As of Talos v1.0` is wrong on the v1.6 through v1.10 Proxmox pages (`proxmox.mdx:121`, and `:93` on v1.6). It should be v1.1.0. Those five files aren't in this diff.

public/talos/v1.10/getting-started/system-requirements.mdx
```diff
@@ -72,3 +72,15 @@ Talos Linux itself only requires less than 100 MB of disk space, but the EPHEMER
 Talos manages disk partitioning automatically during installation, creating EFI, META, STATE, and EPHEMERAL partitions. The EPHEMERAL partition then expands to fill all the space left after the first three. That space can either remain entirely with EPHEMERAL or be divided into additional user volumes, depending on your needs. See [Disk Layout](../configure-your-talos-cluster/storage-and-disk-management/disk-management#disk-layout) for details.
 
 For production, it is often more efficient to dedicate a smaller disk for the Talos installation itself, and use additional disks for workload storage. Using a large, single disk for both system and workloads is supported, but may not be optimal depending on your environment.
+
+## CPU architecture
```
smira: this applies to `amd64` only, `arm64` is not covered here

sterlingkoch (approved):

PR #741, opened 2026-09-03
docs: upgrade Talos docs to v1.14 stable part 3

Iheanacho-ai:
## What
Upgrades the Talos docs to the v1.14.0 stable release, and fixes two bugs in the upgrade tooling that surfaced along the way.

## Why
v1.14.0 is now the latest stable Talos release. While running the upgrade, `TALOS_VERSION` in the Makefile was found stuck on `v1.13` despite `TALOSCTL_IMAGE` already pinned to `v1.14.0` — the version-upgrade generator derived it from an assumed prior value instead of the target tag, so it silently no-op'd once the two drifted. `sync-docs-nav` was also failing outright due to a path typo.

## Change
- `tools/version-upgrade-gen/main.go`: `setMakefilePin` now sets `TALOS_VERSION` unconditionally from the target tag's major.minor via regex, instead of conditionally swapping against an assumed current value — so it can no longer silently drift out of sync with `TALOSCTL_IMAGE`.
- `tools/docs-convert/main.go`: added `document-map.mdx` to the reference-doc generator's preserve-list (alongside `overview.mdx`), so it survives regeneration instead of being deleted.
- `Makefile`: fixed `sync-docs-nav`'s `cd docs-validate` typo (missing `tools/` prefix) that was breaking the upgrade pipeline.
- Regenerated `public/talos/v1.14/reference/**`, `public/changelog.mdx`, and `public/snippets/custom-variables.mdx` (k8s 1.37.0) via the fixed `make upgrade-talos-version-local TAG=v1.14.0`.

## Testing
- Ran: `make broken-links`, `make validate-docs-nav`, `make style-check-changed`, `make docs.json-local` (+ `git diff --exit-code`), `make check-missing-local` — all pass, 0 errors.
- Where: local checkout.
- By: Amarachi, with an agent (Claude) for the tooling fixes and verification.
- Result: all four CI gates plus `check-missing` pass clean; `TALOS_VERSION` correctly tracks `TALOSCTL_IMAGE`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -393,7 +393,7 @@ validate-docs-nav: ## Validate all talos yaml nav configs match their content di
 
 .PHONY: sync-docs-nav
 sync-docs-nav: ## Insert newly generated reference pages into their version YAML nav (best-effort, non-blocking)
-	cd docs-validate && go run . --workspace .. --fix
+	cd tools/docs-validate && go run . --workspace ../.. --fix
 
 # validate-tag distinguishes the two ways a TAG can be wrong, with a tailored
 # message for each, and fails BEFORE the generator writes anything:
--- a/public/changelog.mdx
+++ b/public/changelog.mdx
@@ -4,6 +4,559 @@ description: "Product updates and announcements"
 rss: true
 ---
 
+<Update label="v1.6.1" tags={["Image Factory"]}>
+  [Release notes →](https://github.com/siderolabs/image-factory/releases/tag/v1.6.1)
+
+
+</Update>
+
+<Update label="v1.14.0" tags={["Talos"]}>
+  [Release notes →](https://github.com/siderolabs/talos/releases/tag/v1.14.0)
+
+  ### DNS over TLS (DoT) and DNS over HTTPS (DoH) Support
+
+  Talos now supports DNS over TLS (DoT) and DNS over HTTPS (DoH) for secure DNS resolution.
+  These features allow Talos to encrypt DNS queries and responses, enhancing privacy and security for DNS traffic.
+  The DNS protocol can be configured on a per-name server basis in the `ResolverConfig` document, allowing for flexible configuration of DNS resolution.
+
+  ### Apply Configuration Modes
+
+  The '--mode=reboot' option has been removed from the `talosctl apply-config` command; by default, configuration is applied without a reboot.
+  Most configuration changes don't require a reboot; the documentation lists the changes that do.
+
+  ### Native BGP
+
+  Talos now supports running native BGP routing instances on the host via embedded GoBGP servers, configured with `BGPInstanceConfig` documents.
+  This removes the need to ship FRR as a system extension for the common fabric-facing use case.
+
+  List of changes:
+
+  - Added repeatable, named `BGPInstanceConfig` documents to configure local ASN, router-id, optional Linux VRF, advertised interfaces, neighbors, and per-route preferred source (`routeSource`).
+  - Peer hold-time and BFD behavior are configured inline on each concrete neighbor, which selects either an address or a link.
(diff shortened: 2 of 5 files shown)


sterlingkoch (approved):

sterlingkoch:
Reviewed, nothing blocking. Two notes and a logistics point.

**1. The PR description is missing the biggest change.** [`tools/version-upgrade-gen/main.go:203-213`](https://github.com/siderolabs/docs/blob/5d0e9bf8da679a4690f7f4008c0f40f7647fa4ed/tools/version-upgrade-gen/main.go#L203-L213) and [`:321-383`](https://github.com/siderolabs/docs/blob/5d0e9bf8da679a4690f7f4008c0f40f7647fa4ed/tools/version-upgrade-gen/main.go#L321-L383). `readExportVar` is gone and `k8s_prev_release` now comes from a live GitHub releases lookup, about 75 new lines. Better design than reading back a file the tool just mutated, but it adds a hard network dependency that `os.Exit(1)`s on failure and it's not in the three-item change list. Worth adding so reviewers don't skip past it.

**2. Patch releases will produce a nonsense sample.** [`tools/version-upgrade-gen/main.go:364-383`](https://github.com/siderolabs/docs/blob/5d0e9bf8da679a4690f7f4008c0f40f7647fa4ed/tools/version-upgrade-gen/main.go#L364-L383). `previousStableRelease` takes the release immediately before the tag, so running this for `v1.14.1` picks `v1.14.0` and `k8s_prev_release` ends up equal to `k8s_release`. The only consumer is the illustrative CLI output in `public/kubernetes-guides/advanced-guides/upgrading-kubernetes.mdx:57-63`, which would then read `update kube-apiserver: 1.37.0 -> 1.37.0`. Skip same-minor releases?

**3. Nit, pre-existing.** `public/snippets/custom-variables.mdx:68` still has `release_v1_13 = 'v1.13.7'` while the pointer this PR promotes away from was `v1.13.9`. Now that v1.13 is the previous stable, its docs tell readers to install v1.13.7. Natural moment to bump it.

**4. Logistics.** This is based on `e6ab061` (part 1) rather than current `main`, so part 2 already landed the `TALOS_VERSION := v1.14` fix and a byte-identical reference regen. I test-merged it and it's clean, no revert of part 2, but after a rebase the real diff is just the two Go fixes, the `sync-docs-nav` path, changelog, and custom-variables. Worth rebasing so CI validates the combined state, since it hasn't yet.

Also, [`Makefile:395`](https://github.com/siderolabs/docs/blob/5d0e9bf8da679a4690f7f4008c0f40f7647fa4ed/Makefile#L395) help text says `sync-docs-nav` is "best-effort, non-blocking," but the call sites at `:427` and `:443` are bare `$(MAKE)` with no `-` prefix or `|| true`, so a failure aborts the upgrade. Your own note that the typo "was breaking the upgrade pipeline" is the proof. Either drop "non-blocking" or wrap it the way `validate-docs-nav` does two lines down.

**Verified.** I tested the `TALOS_VERSION` regex against ten line shapes: it matches the real pin, rejects `OMNI_TALOS_VERSION` and `export`-prefixed lines, and `ReplaceAllLiteralString` means a `$` in the replacement can't expand. `minorOf` handles `v1.14.0`, `v1.14.0-beta.0`, and `v1.15.0-alpha.1` correctly, and major.minor is right since it's consumed as a folder path. The `document-map.mdx` preserve entry matches on exact filename via `filepath.Base`, so no over-matching, and the file really was inside the delete-walk's `dstDir` and is referenced at `talos-v1.14.yaml:251`, so this is a real save. `tools/docs-validate/` exists and `--workspace ../..` is the established form. Zero deletions or renames in the diff, so no dangling-nav risk. All `docker run` calls are `--rm` with no `-it`.

PR #749, opened 2026-09-10
docs: update EIF overview page and add TEL introduction

### LINKED ISSUE #652: Update the EIF overview page
Part of #649. **Partially unblocked as of September 3.** Two of the three edits can be written now; the third stays blocked.

The [EIF page](https://docs.siderolabs.com/talos/v1.13/learn-more/enterprise-image-factory) is the customer-facing explanation of the enterprise factory, and it has drifted from the September product in three places. This edit lives in the Talos tree and does not wait on #677.

**Writable now:**

1. **The tier list is retired.** The page still advertises Enterprise / Edge / FIPS tiers. September's structure is a single Talos Enterprise Linux entitlement, and it includes Omni. There is no separate FIPS tier to advertise.

2. **The FIPS claim needs its certificate.** Replace loose FIPS wording with: FIPS-enabled builds using the Go Cryptographic Module, FIPS 140-3, [CMVP certificate #5247](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/5247). Never write that Talos is FIPS-certified. Validation attaches to the cryptographic module, not to the OS as a whole, which is why the claim is always phrased this way.

**Still blocked:**

3. **The two customer-provided-extension references**, one in the SBOM attribution and one in the on-prem description of reading customer-private content from the customer's own registry. There is a mechanism now (an additive extension catalog, `enterprise.extraExtensions`, shipped in image-factory v1.4.0), but it is configured once per factory deployment by whoever operates the factory rather than per customer, and no customer can reach it today. What the page should say about it is an open product question. Leave both references as they are for now.

Diátaxis type: explanation (update).

Iheanacho-ai:
## What

Updates the Enterprise Image Factory overview page (v1.13 and v1.14) and adds an introductory TEL overview page.

## Why

Closes #652.

## Change

- `public/talos/v1.13/learn-more/enterprise-image-factory.mdx` and `v1.14/...` — retired the tier list (Enterprise / Edge / FIPS); replaced with single-entitlement language; updated FIPS section to name the Go Cryptographic Module and CMVP certificate #5247; removed the retired FIPS product tier reference
- `public/tel/how-to-guides/overview.mdx` — new introductory page covering what TEL is, included features (FIPS, SBOM, VEX, scan reports), and access
- `tel.yaml` — registers the overview page in TEL navigation
- `Makefile`, `public/docs.json` — updated and regenerated

## Testing

- Ran: `make broken-links`, `make style-check-changed`, `make docs.json && git diff --exit-code`
- By: agent (Claude)
- Result: 0 errors, 0 warnings on changed files; pre-existing warnings on unchanged headings in the EIF pages
-


public/talos/v1.14/learn-more/enterprise-image-factory.mdx
```diff
@@ -10,9 +10,9 @@ Though the Enterprise Image Factory ( `factory.siderolabs.com` ) and the public
 - `factory.talos.dev` is anonymous and unauthenticated. It is the distribution mechanism for open-source Talos Linux.
 - `factory.siderolabs.com` requires authentication and is the delivery channel for Talos Enterprise Linux artifacts: FIPS-enabled builds, per-schematic SBOMs, and VEX data.
 
-## Tier availability
+## Availability
 
-Talos Enterprise Linux is available to customers on the Enterprise, Enterprise On-Prem, Edge, Edge On-Prem, and Talos Enterprise Linux Support tiers. It is also bundled into the FIPS product, a separate tier for customers who need FIPS 140-3 (CMVP) certified builds, on top of everything included in Talos Enterprise Linux. Customers on the Talos Enterprise Linux Support tier authenticate to Enterprise Image Factory directly, without an Omni account.
+Talos Enterprise Linux is a single entitlement that includes Omni. Customers on the Talos Enterprise Linux Support tier, who do not use Omni, authenticate to Enterprise Image Factory directly.
```
utkuozdemir: The tiers, "a single entitlement that includes Omni", and the FIPS wording are product statements, and this version changes them: today's page says FIPS-enabled builds are distinct from CMVP certification, this one says "validated under FIPS 140-3 (CMVP certificate #5247)", and the FIPS tier is gone. The certificate number itself is right, it is the Go cryptographic module's. Someone from product should confirm the rest, I cannot. @sterlingkoch?

public/talos-enterprise-linux/overview/introducing-talos-enterprise-linux.mdx
```diff
@@ -0,0 +1,26 @@
+---
+title: Introducing Talos Enterprise Linux
+description: An overview of Talos Enterprise Linux, what it includes, and how to get started.
+---
+
+import { version } from '/snippets/custom-variables.mdx';
+
+Talos Enterprise Linux is Sidero's commercial distribution of Talos Linux, delivered through Enterprise Image Factory at `factory.siderolabs.com`. It adds compliance and supply chain artifacts — FIPS-enabled builds, per-schematic SBOMs, VEX data, and vulnerability scan reports — that are not available in the open-source distribution.
```
utkuozdemir, line 8: The existing docs use "public Image Factory", "Enterprise Image Factory" and "Talos Enterprise Linux" spelled out. The new pages also use "community", "open-source", "enterprise factory", "the factory" and "TEL". I could suggest settling on the existing terms and using them everywhere, no acronyms. @sterlingkoch, what do you think?

public/tel/how-to-guides/overview.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+- **VEX data:** Sidero publishes Vulnerability Exploitability eXchange statements that identify which CVEs actually apply to Talos as built. Used with the SBOM, VEX data lets scanners filter out non-applicable findings rather than relying on blanket waivers.
+
+- **Vulnerability scan reports:** Enterprise Image Factory can run and return a completed vulnerability scan report on demand, combining the schematic's SBOM with VEX data and a live vulnerability database. Reports are available in `.json`, `.table`, `.sarif`, and `.cdx` formats.
+
+For a full breakdown of features and how Enterprise Image Factory differs from the public Image Factory at `factory.talos.dev`, see <a href={`../../../talos/${version}/learn-more/enterprise-image-factory`}>Enterprise Image Factory</a>.
+
+## Access
+
+Talos Enterprise Linux is included with qualifying Omni subscriptions and is also available as a standalone entitlement. Customers on the Talos Enterprise Linux Support tier, who do not use Omni, authenticate to Enterprise Image Factory directly.
```
sterlingkoch: Two problems in one sentence. "Included with qualifying Omni subscriptions" is backwards, Omni is part of Talos Enterprise Linux rather than the other way round. And "Talos Enterprise Linux Support tier" is retired tier language, same as on the EIF page.

public/talos/v1.14/learn-more/enterprise-image-factory.mdx
sterlingkoch: Product side, taking all three.

**Tiers.** Removing the list is right, access isn't gated per tier anymore. The replacement still carries tier language in four places on this page though:

:16 reintroduces "Talos Enterprise Linux Support tier," which is also retired
:25 fixed the values but the row label still reads "Tier access"
:96 "Omni-managed customers on a qualifying tier"
:110 "does not require an Omni account on the Talos Enterprise Linux Support tier"
Same four in the v1.13 copy. All of these should describe the customer rather than a tier.

**"A single entitlement that includes Omni."** Correct. Omni is part of Talos Enterprise Linux rather than a separate purchase. Note overview.mdx:26 states it the other way round, "included with qualifying Omni subscriptions," and needs flipping.

**FIPS.** The new wording is better than what's on the page today, and :45 gets the load-bearing part right: the validation attaches to the Go Cryptographic Module, not to Talos Linux. Keep it phrased that way everywhere, and note the certificate is Geomys', not ours, which is the reason it can't attach to the OS.

public/tel/how-to-guides/overview.mdx
sterlingkoch: Same comments on two other PRs:
* download-image-and-artifacts.mdx:18 in #751 says "a qualifying Talos Enterprise Linux tier,"  
* manage-credentials.mdx:36 in #753 says "Talos Enterprise Linux Support tier users."

public/talos-enterprise-linux/overview/introducing-talos-enterprise-linux.mdx
sterlingkoch: Agreed, and let's make it a rule rather than a preference. Full names everywhere: Talos Enterprise Linux, Enterprise Image Factory, public Image Factory. No "TEL", no "enterprise factory", no bare "the factory". Where we mean the free product it's "open-source Talos Linux", which is what the rest of the docs already say; "community Talos" is new vocabulary and I'd rather not introduce a second word for the same thing.

For scale, the eight new pages use "the factory" 16 times against 4 in the entire existing corpus, so this is worth fixing now rather than after publish.

One consequence worth deciding deliberately: the acronym is also in the URL path (tel/...) and in three filenames (scan-tel-with-sbom-and-vex, getting-started-with-tel, migrate-clusters-to-tel). If the rule is no acronyms, those should change too, and it's a lot cheaper before these pages are live and indexed.

tel.yaml
```diff
@@ -0,0 +1,10 @@
+navigation:
+  version: "v1.0"
```
sterlingkoch: This gives the tab a version dropdown but the pages underneath have no version segment, so there's one entry over a tree that can't hold a second version. Two ways to go:

Implement the versioning now: pages move to tel/v1.14/... and this becomes version: "v1.14", mirroring talos-v1.14.yaml.
Drop the version: key and let the TEL tab be unversioned like Omni, and revisit when there's a reason.
I want option 1, since the TEL section tracks Talos versions. What I don't want is the current state, where the label says versioned and the paths say otherwise.

If we're also dropping the tel acronym from the path, do both in the same move so it's one rename of eight files rather than two.

tel.yaml
```diff
@@ -0,0 +1,10 @@
...
+  tabs:
+    - tab: "Talos Enterprise Linux"
+      icon: "/images/talos.svg"
+      groups:
+        - group: "How to guides"
+          folder: "tel/how-to-guides"
+          pages:
+            - "tel/how-to-guides/overview.mdx"
```
sterlingkoch: This registers overview and nothing else. The other seven pages across #751, #753, #754, #756, #757 and #759 have no nav entry, so they'd build and publish unreachable. I checked by applying all seven together and regenerating: the whole Talos Enterprise Linux tab comes out as one page.

From Claude:
Worth knowing why the checks didn't catch it. make validate-docs-nav only inspects trees it finds in a YAML file, and there's no tel.yaml on main yet, so it passes on each page PR in isolation. The moment this merges, the other six start failing it. So each of those PRs needs its own line added here as part of landing.

This file also needs a learn-more group for api-tokens.mdx from #753, assuming that page survives your thread about merging it into the how-to. That one is the single orphan validate-docs-nav can't see, because it has no declared folder to compare against.

sterlingkoch (commented):
One blocker that isn't anyone's PR yet. tools/frontmatter-check requires every tree under public/ to be registered before it can pass, and public/tel isn't:

```
$ cd tools/frontmatter-check && go run . ../../public/tel
1 file(s) belong to no known section: ../../public/tel
A new content tree must be registered before it can pass this check.
exit status 1
```
Fix is one line in the sections list at main.go:42, alongside the Omni entry:

`{dir: "public/tel", fields: []string{"title", "description"}},`

Title and description, not canonical, which is Talos-tree only. None of these seven PRs have hit this because the step wasn't in docs-ci.yaml when they last ran and the failing ones die earlier at broken-links. It's on main now, so every one of them fails this on re-run.

tel.yaml
Iheanacho-ai: I purposefully left out all the files from the tel.yaml, and i will delete this tel.yaml and the changes it makes from the docs.json and have a single pr that creates the structure @sterlingkoch

tel.yaml
Iheanacho-ai: deleted the structure

sterlingkoch (approved):

Iheanacho-ai:
superseded by #771

PR #750, opened 2026-09-10
fix for docker run command syntax code & arm64 docs support for omni

hegerdes:
## What
Two commits with two changes:

 * Fix docker run code blocks syntax for talos
 * Support ARM64 for Omni on-prem docs 

## Why
 * Docker run bash command is invalid
 * Omni and `CFSSL` fully support ARM - so docs should handle this

## Change
Code instructions with fixes or additions

## Testing
None - are there tests for your docs?

## Notes
If requested I can separate the change in two PRs


public/talos/v1.7/platform-specific-installations/local-platforms/docker.mdx
```diff
@@ -124,22 +124,22 @@ kubectl --context admin@cluster2 get nodes
...
-  --mount type=volume,destination=/system/state
-  --mount type=volume,destination=/var
-  --mount type=volume,destination=/etc/cni
-  --mount type=volume,destination=/etc/kubernetes
-  --mount type=volume,destination=/usr/libexec/kubernetes
-  --mount type=volume,destination=/opt 
-  -e PLATFORM=container 
+docker run --rm -it \
+  --name tutorial \
+  --hostname talos-cp \
```
sterlingkoch: These CodeBlocks are JS template literals, so a single trailing \ is a line continuation and gets eaten along with the newline. The block renders as one long line. Double them (\\), same as upgrading-talos.mdx:108. Applies to all nine files.

sterlingkoch (commented):
No need to split. The cfssl arm64 change is good as-is, asset names check out for linux/aarch64. One fix needed on the docker blocks, see inline. Also heads up that #777 deletes the v1.6 page you edited here, so drop that file; v1.7 survives but is leaving the version switcher.

hegerdes:
Done. Dropped 1.6 change, added escaped backslash and `lang="sh"` to codeblocks

sterlingkoch (approved):

PR #751, opened 2026-09-10
docs: add Enterprise Image Factory download guide

### LINKED ISSUE #650: How-to: Download images and artifacts from Enterprise Image Factory

Iheanacho-ai:
## What

Adds `public/tel/how-to-guides/download-image-and-artifacts.mdx`, covering how to
download Talos Enterprise Linux boot assets, SBOMs, VEX data, and vulnerability
scan reports from Enterprise Image Factory, through the UI and programmatically.

Fixes #650 

## Why

Enterprise Image Factory serves artifacts the public Image Factory does not —
per-schematic SBOMs, per-release VEX, and scan reports — and every request is
authenticated. None of that was documented.

## Change

New page with:
- UI wizard flow, and a note that schematic IDs are account-scoped
- Boot asset URL structure and the common asset path table
- SBOM (`/spdx`), VEX (`/vex`), and scan report (`/scans`) endpoints, including
  the four report formats
- Checksum verification, scoped to boot assets
- Scope requirements: `report:read` for the artifact endpoints,
  `schematic:create` for `POST /schematics`
- `-L` on every boot asset example, since those requests redirect

## Testing

- **Ran:** every command in the programmatic section, by hand, against
  `factory-enterprise.staging.talos.dev` with Talos v1.14.0, amd64, schematic
  `551f50cf…898d`.
- **By:** me, with Claude driving the test plan.
- **Result:** all endpoints and claims verified — SPDX 2.3 confirmed in the
  payload, VEX confirmed as OpenVEX v0.2.0 and per-release, all four report
  suffixes (`.json`/`.table`/`.sarif`/`.cdx`) return 200, `.sha256`/`.sha512`
  return checksums, all eight asset paths resolve, unauthenticated requests 401,
  and `POST /schematics` returns 201.
- **Fixed during testing:**
  - Boot assets return 302 to a pre-signed storage URL. Without `-L`, curl saves
    the redirect page — a 537-byte HTML fragment written to a file named like the
    image, which looks like a successful download. Added `-L` and `-o`.
  - `.sha256`/`.sha512` 400/404 on the artifact endpoints, so "any asset path"
    was wrong; narrowed to boot assets.
  - A `Talos` actor token (`image:read` only) 401s on every artifact endpoint.
    Documented the required scopes.


public/tel/how-to-guides/download-image-and-artifacts.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+Appending `.sha256` or `.sha512` to any boot asset path returns a checksum file instead of the asset:
+
+For example to download the checksum file of the asset path `metal-amd64.raw.zst`:
+
+```bash
+curl -L -H "Authorization: Bearer <token>" https://factory.siderolabs.com/image/<schematic>/<version>/metal-amd64.raw.zst.sha256
+```
+
+The checksum verifies the integrity of the asset as served to you — that your download arrived intact. Because some images are generated dynamically and are not byte-for-byte reproducible across builds, checksums are not stable identifiers to pin over time. For cryptographic verification of provenance, use signatures instead.
```
utkuozdemir: Where does this come from? I could not find it in the factory docs, and as far as I know the checksum is just a plain `sha256sum -c` check. If there is no source, I would cut the sentence and keep the checksum steps.

public/talos-enterprise-linux/working-with-talos-enterprise-linux/download-images-and-artifacts.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+If you have used the public [Image Factory](https://factory.talos.dev/) or the **Create New Media** wizard in Omni, this is the same flow, the only difference is that you authenticate first.
+
+To download boot assets from the UI:
+
+1. Log in to [Enterprise Image Factory](https://factory.siderolabs.com/).
+
+2. On each page of the wizard, select the appropriate options for your setup, then click **Next** to continue.
+
+   The wizard is split across multiple pages. Each page presents a different set of configuration options — such as hardware type, architecture, Talos Linux version, and system extensions — that you can use to customize your Talos Enterprise Linux image.
```
utkuozdemir, line 32: I think this can be one line, the reader knows what a wizard is.

public/tel/how-to-guides/download-image-and-artifacts.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+## Download boot assets programmatically
+
+Boot assets follow the same URL structure as the public Image Factory, served from `factory.siderolabs.com`:
+
+```text
+https://factory.siderolabs.com/image/<schematic>/<version>/<asset-path>
+```
+
+However, unlike the public Image Factory, the Talos Enterprise Image Factory requires every request to be authenticated with a token. To learn how to obtain your token, see [Manage Enterprise Image Factory Credentials](./manage-credentials).
```
utkuozdemir: "Enterprise Image Factory", like everywhere else.

sterlingkoch (dismissed):

public/tel/how-to-guides/download-image-and-artifacts.mdx
sterlingkoch: Cutting it is right, there's no source for it. Worth knowing the same claim shows up inverted in #759:123, which says the schematic ID plus version and architecture "uniquely identifies a reproducible image". This page says images are generated dynamically and aren't byte-for-byte reproducible. We can't support either, so I'd cut both and let the checksum steps speak for themselves.

public/talos-enterprise-linux/working-with-talos-enterprise-linux/download-images-and-artifacts.mdx
Iheanacho-ai: i think it would be better to just keep the line as is, as we are not sure what page a user lands first. whether that be the omni docs or the tel docs. basically the same reason to keep this small description in the omni docs, is the same reason to keep it in tel

sterlingkoch (approved):

Iheanacho-ai:
superseded by #771

PR #753, opened 2026-09-10
docs: add Enterprise Image Factory authentication pages

Iheanacho-ai:
## What

Two new pages for Talos Enterprise Linux:

- `tel/how-to-guides/manage-credentials` — how to create, use, rotate, and revoke API tokens
- `tel/learn-more/api-tokens` — concept reference for token scopes, actor profiles, scope delegation, ownership, and lifecycle

Also adds `tel.yaml` for TEL navigation and regenerates `docs.json`.

## Why

Closes #653.

## Change

- `public/tel/how-to-guides/manage-credentials.mdx` — new how-to covering machine configuration patching, OCI/registry auth, Bearer header, and `?token=` URL parameter usage
- `public/tel/learn-more/api-tokens.mdx` — new concept page covering all eight scopes, four actor profiles, scope delegation, token ownership, and stored vs. ephemeral token lifecycle
- `tel.yaml` — new nav config for the TEL section
- `Makefile` — updated to include `tel.yaml` in the nav build targets
- `public/docs.json` — regenerated

## Testing

- Ran: `make broken-links`, `make validate-docs-nav`, `make style-check-changed`, `make docs.json && git diff --exit-code`
- By: agent (Claude)
- Result: all checks pass, 0 errors, 0 warnings
- Filled from outside the page: token mechanics sourced from `image-factory` repo (`docs/authentication.md`, `docs/api.md`, `internal/frontend/http/templates/tokens.html`) and a call transcript with the implementing engineer

Issue #653, opened 2026-07-16
How-to: Get and manage Enterprise Image Factory credentials


sterlingkoch:
Part of #649. **Drafted in [docs#753](https://github.com/siderolabs/docs/pull/753).** Do not restart this page.

The token how-to and the API-tokens concept page are in that PR. Token mechanics (actor profiles, Basic vs Bearer, machine-config patch, rotation, revocation) stand.

## What to change on #753, not rewrite

Planning Marathon on September 9 retracted the "get an organization" how-to. Public docs start after the customer can already log in. Organization creation is sales and ops (Terraform at contract close). It is not a customer procedure and it is not "email Spencer" or a support form.

On `manage-credentials.mdx`:

- Drop the "Get access to an organization" subsection as a how-to. A one-liner is enough if you need a prerequisite at all: your organization exists once the contract is in place. Omni users never request one.
- Do not document who to contact, and do not describe the internal provisioning process.
- Do not use "Talos Enterprise Linux Support tier". That name is dead.
- Don't describe identity-provider sign-in as current until production has cut over. Production was still on the old factory auth on September 4.

The rest of the page (create / use / rotate / revoke) can stay.

## Still open, publishing not drafting

Which release production cuts over to, and whether existing htpasswd credentials keep working through an overlap window. Don't describe the sign-in flow as current until it is.

Diátaxis type: how-to. The concept page in #753 is the reference sibling.

PR #757, opened 2026-09-11
docs: add docs on how to migrate to enterprise linux talos 

### LINKED ISSUE #656: Guide: Migrating from community Talos to Talos Enterprise Linux
Part of #649. **Unblocked. Nothing drafted yet.**

The migration guide for existing Talos users moving to TEL.

**Launch order, from Planning Marathon on September 9:** SaaS Omni first, on-prem second. A SaaS customer does not walk a four-step talosctl path. Once the account is on the enterprise factory, they trigger an upgrade.

Scope:

- How to move clusters and machines to enterprise images.
- Default path: SaaS Omni. The account is feature-flagged onto the enterprise factory, then the operator upgrades. Do not invent extra steps in front of that.
- Secondary path: talosctl-only / self-hosted Omni pointing at the hosted enterprise factory. Technical, not the default. Utku is drafting that as a PR; Amarachi places it. Don't invent the factory-config steps on this page while that PR is coming.
- The one-way door, stated plainly. Moving to enterprise images is seamless: both factories sign with the same keys, so nothing on the machine breaks at the boundary, including SecureBoot-enrolled nodes. Omni does not downgrade a machine from enterprise back to community images. That boundary is product policy and support posture, not a technical lock. Running clusters are unaffected if you leave, but enterprise upgrades to enterprise.
- What changes operationally (image sources, upgrade flows) and what doesn't (the OS is the same Talos).

Do not document how to get a factory organization. Access is provisioned at contract close. This page starts after the customer can already reach the factory. Token mechanics live on #653 (drafted in #753).

Diátaxis type: how-to + explanation.

Iheanacho-ai:
## What

New how-to page: `tel/how-to-guides/migrate-clusters-to-tel`

Covers both migration paths for moving existing Talos nodes to Talos Enterprise Linux:
- **SaaS Omni** — automatic switch on the next cluster upgrade once the account is provisioned on Enterprise Image Factory
- **talosctl** — patch machine configuration with factory credentials, then run `talosctl upgrade` with an enterprise installer image

Includes a verification step (`talosctl get version --output yaml`, check `spec.name`).

## Why

Closes #656.

## Change

- `public/tel/how-to-guides/migrate-clusters-to-tel.mdx` — new how-to

The credential setup and token details referenced in the talosctl path are covered by the `manage-credentials` page (#653 / PR #753) and linked from this page.

## Testing

- Ran: `make broken-links`, `make validate-docs-nav`, `make style-check-changed`, `make docs.json && git diff --exit-code`
- By: me
- Result: all checks pass
- Filled from outside the page: SaaS Omni behavior (auto-switch on next schematic change, Talos 1.13+ threshold, boot-source variations), bare-talos upgrade flow, and version verification method sourced from call with Utku


public/tel/how-to-guides/migrate-clusters-to-tel.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+The patch sits in the machine configuration and does nothing until an upgrade is triggered.
+
+### Step 2: Run the upgrade
+
+Run `talosctl upgrade` and pass an enterprise installer image from `factory.siderolabs.com`:
+
+```bash
+talosctl upgrade --nodes <node-IP> \
+  --image factory.siderolabs.com/installer/<schematic-id>:<talos-version>
```
utkuozdemir: Since Talos 1.10 the installer image names carry the platform, e.g. `factory.siderolabs.com/metal-installer/<schematic-id>:<talos-version>`, and that is what the factory UI and Omni use. The plain `installer` name still works as the legacy name and resolves to the metal variant, so this is not wrong, but I would use `metal-installer` here to match the UI.

public/talos-enterprise-linux/working-with-talos-enterprise-linux/migrate-clusters.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+Installation media downloaded through the Omni UI or `omnictl download` always comes from your provisioned factory and is automatically enterprise.
+
+Disk images, such as AWS AMIs and pre-installed metal images, downloaded through Omni are already enterprise on first boot. Community ISOs initially appear as non-enterprise in the machine list, but when you create a cluster from one, Omni pulls the enterprise installer, installs it to disk, and reboots the node into enterprise Talos.
+
+## Migrate with talosctl
+
+If you manage Talos machines without Omni, migration is a two-step process: add your Enterprise Image Factory credentials to the machine configuration, then run an upgrade targeting an enterprise installer image.
+
+### Step 1: Add credentials to the machine configuration
```
utkuozdemir, line 28: The `RegistryAuthConfig` document is supported from Talos 1.12, and enterprise images exist from 1.13. So the node needs to be on 1.12 or later before this step. Maybe worth a prerequisite line?

public/talos-enterprise-linux/working-with-talos-enterprise-linux/migrate-clusters.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+```bash
+talosctl patch machineconfig --nodes <node-IP> --patch @credentials.yaml
+```
+
+The patch sits in the machine configuration and does nothing until an upgrade is triggered.
+
+### Step 2: Run the upgrade
+
+Run `talosctl upgrade` and pass an enterprise installer image from `factory.siderolabs.com`:
```
utkuozdemir, line 40: The schematic ID needs to come from the enterprise factory, the IDs from the public factory do not exist there (the owner is part of the ID). Maybe one sentence pointing to the wizard at `factory.siderolabs.com`, or to the download guide?

public/tel/how-to-guides/migrate-clusters-to-tel.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+Migrating an existing Talos cluster to Talos Enterprise Linux is an in-place upgrade. The migration changes the image source from the public factory at `factory.talos.dev` to the Enterprise Image Factory at `factory.siderolabs.com`, while preserving the existing machine configuration and cluster state.
+
+This guide explains how to migrate Talos Linux and Omni-managed clusters and verify that a node is running Talos Enterprise Linux.
+
+You must have a valid Enterprise Image Factory account. If you cannot log in, contact [support](https://www.siderolabs.com/support).
+
+## Migrate with SaaS Omni
+
+Once your account is provisioned on the enterprise factory, Omni switches each cluster to enterprise images on its next upgrade. No manual action is required beyond triggering an upgrade; any schematic change will do, including bumping the Talos version, adding or removing a system extension, or changing kernel arguments. Until the cluster is upgraded, its machines continue running community Talos.
```
utkuozdemir: Small correction: Omni does not switch anything by itself, the next Talos upgrade does. Something like: "Existing clusters keep running community Talos. A cluster moves to enterprise on its next Talos upgrade, once it runs 1.13 or later."

Also, which subscriptions get the enterprise factory and when is a product topic, and it may change. I would leave "Once your account is provisioned on the enterprise factory" out of the technical docs and start the section from the clusters. @sterlingkoch, what do you think?

public/talos-enterprise-linux/working-with-talos-enterprise-linux/migrate-clusters.mdx
```diff
@@ -0,0 +1,57 @@
+---
+title: Migrate Clusters to Talos Enterprise Linux
+description: How to move existing Talos clusters and machines to Talos Enterprise Linux images through SaaS Omni or talosctl.
+---
+
+Migrating an existing Talos cluster to Talos Enterprise Linux is an in-place upgrade. The migration changes the image source from the public factory at `factory.talos.dev` to the Enterprise Image Factory at `factory.siderolabs.com`, while preserving the existing machine configuration and cluster state.
+
+This guide explains how to migrate Talos Linux and Omni-managed clusters and verify that a node is running Talos Enterprise Linux.
+
+You must have a valid Enterprise Image Factory account. If you cannot log in, contact [support](https://www.siderolabs.com/support).
```
utkuozdemir, line 10: I think only the `talosctl` section needs a factory login, Omni users need nothing. Maybe move it there?

Iheanacho-ai: But not all Omni users can access Enterprise Image Factory can they? I think a valid account whether or not you are using Omni still stands

public/talos-enterprise-linux/how-to-guides/migrate-clusters.mdx
```diff
@@ -0,0 +1,61 @@
...
+```bash
+talosctl upgrade --nodes <node-IP> \
+  --image factory.siderolabs.com/metal-installer/<schematic-id>:<talos-version>
+```
+
+Talos uses the credentials in the machine configuration to authenticate against Enterprise Image Factory, pulls the installer image, installs it, and reboots the node into Talos Enterprise Linux.
+
+Repeat for each node in the cluster.
+
+## Verify a node is running Talos Enterprise
```
sterlingkoch: Should be "Talos Enterprise Linux"

public/talos-enterprise-linux/working-with-talos-enterprise-linux/migrate-clusters.mdx
sterlingkoch: Utku's point and your reply are both partly right, and the fix is one word. An Omni user with a TEL entitlement does need the entitlement, which is what you're defending, but never needs to log into Enterprise Image Factory, which is what he's pointing at. Suggest: require the entitlement up top, and move "if you cannot log in, contact support" down into the talosctl section, where logging in is actually a step.

sterlingkoch (commented):
omnictl download is deprecated as of v1.12.0, replaced by omnictl media download. See #765.

public/talos-enterprise-linux/working-with-talos-enterprise-linux/migrate-clusters.mdx
```diff
@@ -0,0 +1,61 @@
...
+Talos uses the credentials in the machine configuration to authenticate against Enterprise Image Factory, pulls the installer image, installs it, and reboots the node into Talos Enterprise Linux.
+
+Repeat for each node in the cluster.
+
+## Verify a node is running Talos Enterprise Linux
+
+After an upgrade, confirm a node is running Talos Enterprise Linux by checking its version name:
+
+```bash
+talosctl version --nodes <node-IP>
```
sterlingkoch: talosctl version doesn't include the name, so this command shows nothing to check. The server block is rendered from machine.VersionInfo, which only carries Tag, SHA, Built, Go version and OS/Arch, and versionTemplate prints exactly those plus NODE and Enabled. There's no Name field on that path.

The name lives on a different surface: the COSI resource Versions.runtime.talos.dev, whose spec is {name, version}. That's what talosctl get version --output yaml read in the previous version of this page, and it's the same resource Omni polls to set the enterprise machine label. Suggest going back to that command.

Line 61 needs a fix either way. "If the field is absent, the node is still running open-source Talos Linux" isn't right: on the COSI resource, community builds report Talos rather than nothing, since that's the build default. Absence means you're reading the wrong resource, not that you're on a community node. The previous wording, Talos versus Talos Enterprise, was correct.

One thing to check on your instance since you have one: the resource type is Versions.runtime.talos.dev, so confirm whether the short talosctl get version alias resolves, or whether the page should spell out the fully qualified name.

public/talos-enterprise-linux/working-with-talos-enterprise-linux/migrate-clusters.mdx
sterlingkoch: Still open on the current head, line 10. Flagging in case it got lost under the other changes.

public/talos-enterprise-linux/working-with-talos-enterprise-linux/migrate-clusters.mdx
Iheanacho-ai: oh yeah the changes reverted

public/talos-enterprise-linux/working-with-talos-enterprise-linux/migrate-clusters.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+## Verify a node is running Talos Enterprise Linux
+
+After an upgrade, confirm a node is running Talos Enterprise Linux by checking its version name:
+
+```bash
+talosctl get version --nodes <node-IP> --output yaml
+```
+
+In the output, look for the `name` field in the spec. `Talos Enterprise` means the node is running Talos Enterprise Linux.
```
sterlingkoch: Can we put the community case back? "Talos means it is still on an open-source Talos Linux image" was right, and without it the reader sees name: Talos and has to infer what that means.

sterlingkoch (approved):

Iheanacho-ai:
superseded by #771

PR #760, opened 2026-09-11
docs: put the Multus host-local IPAM store on tmpfs

daixtrose:
## What

Adds `"dataDir": "/run/cni"` to the `host-local` IPAM block of the `NetworkAttachmentDefinition` example on the **Deploy Multus CNI** page, with two short paragraphs after it: one saying why that path is the right one on Talos, one on the node-local scope of `host-local`.

Page: https://docs.siderolabs.com/kubernetes-guides/cni/multus — "Configure a multi-homed pod", step 2. Source: `public/kubernetes-guides/cni/multus.mdx`. The Kubernetes guides are unversioned, so this is the only copy of the page; no backport to versioned Talos directories applies.

## Why

`host-local` records each allocation as a file under `/var/lib/cni/networks/$NETWORK_NAME`, which is on the persistent partition and survives a reboot. A reboot may not deliver a CNI `DEL` for the pods that were running, so addresses can stay allocated after those pods are gone and the usable range shrinks over time.

The plugin documents the remedy itself — [host-local](https://www.cni.dev/plugins/current/ipam/host-local/): *"Allocated IP addresses are stored as files in `/var/lib/cni/networks/$NETWORK_NAME`. The path can be customized with the `dataDir` option …"*, and it points environments where IPs are released automatically on reboot (running containers are not restored) at a tmpfs mounted directory instead. Talos is exactly such an environment.

Two facts make `/run/cni` the right path here:

- `/run` is a tmpfs on Talos, so the store starts empty on every boot. The mount is in the Talos source and has been for every supported release: `internal/pkg/mount/v3/helpers.go` on `main` and in v1.13.7, `internal/pkg/mount/v2/pseudo.go` in v1.9.0, `internal/pkg/mount/pseudo.go` in v1.6.0, each mounting `tmpfs` at `/run`; the [Talos architecture page](https://docs.siderolabs.com/talos/v1.13/learn-more/architecture) says the same, and `talosctl mounts` on the node confirms it. This holds whether or not `/var/run` is a symlink to `/run` on a given version, so it is independent of the version check earlier on the page.
- The Multus thick daemon chroots delegate plugins into the host root — `"chrootDir": "/hostroot"` in `multus-daemon-config`, with `kube-multus-ds` mounting hostPath `/` — so the directory `host-local` writes to is the node's `/run` and not a path inside the Multus pod.

## Change

One file, six added lines: `"dataDir": "/run/cni"` added to the example's `ipam` block, and two explanatory paragraphs after the block — one on `dataDir`, one on the node-local scope of `host-local` — in the page's existing voice. No nav change, no new page, no change to the Known issues section — this is a defect in the example itself rather than an open upstream bug with a workaround.

## Testing

Measured on Talos v1.13.9 / Kubernetes v1.36.4, with a deliberately single-address `host-local` range so that one pod is enough to exhaust it.

Before the change: the pod holding the single address went to phase `Failed` across a node reboot while still carrying that address in its `k8s.v1.cni.cncf.io/network-status` annotation, and every replacement pod the Deployment created afterwards came up with no `network-status` annotation at all, because the range had no free address left. The service stayed without an address for 28 minutes across four further reboots, until the stranded pod was deleted by hand. One reboot took the only address in the range out of circulation for the rest of the series.

After the change, on the same node immediately after the rollout, the live store is on tmpfs and the persistent file is gone. Node name and address are redacted and the network is shown as `my-net`:

```text
talosctl ls -l -r -n <node> /run/cni
drwxr-xr-x  run_t  my-net
-rw-------  run_t  my-net/<addr>
-rw-------  run_t  my-net/last_reserved_ip.0

talosctl ls -l -n <node> /var/lib/cni/networks/my-net
-rw-------  cni_state_t  last_reserved_ip.0
```

The lease file that had survived the reboot is no longer under `/var/lib`; the live one sits under `/run`, which the kernel clears on every boot.

Repository checks, run locally on this branch:

- `make broken-links` — `success no broken links found`.
- `make validate-docs-nav` — all versions OK.
- `make check-frontmatter-changed` on the changed file — `Checked 1 file(s): 0 issue(s)`.
- Style checker on the changed file — `0 error(s), 3 warning(s)`; all three warnings are pre-existing `headings/sentence-case` hits on headings containing "Multus" (lines 19, 275, 285) that this change does not touch.
- `make docs.json` — the regenerated file is byte-identical to the committed one, as expected for a change that touches no nav YAML.
- Environment: the Go tools were run in a `golang:1.26` container and the link check in `ghcr.io/siderolabs/docs-mint:latest`. `ghcr.io/siderolabs/style-guide-checker:latest` is not pullable anonymously (`denied`), so the checker was built from `tools/style-guide-checker` instead, which is the `-local` path the Makefile and `CLAUDE.md` document as equivalent.
- Filled from outside the page: that `host-local`'s default store is persistent, that `/run` is a tmpfs on Talos, and that the Multus thick daemon chroots into the host root. None of the three is stated on this page today; the first is in the plugin's own documentation, the second in the Talos source and the architecture page, the third in the Multus thick DaemonSet's own config.
- Not run: `make preview`, which needs the Mintlify preview server; the change is six lines of MDX with no new component or frontmatter field.

The analysis and this change were prepared with the help of Claude (Anthropic) and reviewed and tested by me; I take responsibility for the content and am happy to rework any part of it.


daixtrose:
Thanks — both warnings in the automated review are correct, and de76d37 addresses them.

**On the CNI `DEL` claim.** You are right that the paragraph overstated it. I checked containerd 2.2.7, the version Talos v1.13 ships: `recover` rebuilds sandboxes from the metadata under `/var/lib/containerd`, which survives a reboot, and `getNetNS` is a plain `netns.LoadNetNS(meta.NetNSPath)` with no I/O, so `sandbox.NetNS` is non-nil for any non-host-network pod even though the netns bind mount under `/run/netns` is gone. `stopPodSandbox` then takes the `if sandbox.NetNS != nil` branch, sets `NetNSPath` to `""` because `Closed()` reports true, and calls `teardownPodNetwork` anyway; `RemovePodSandbox` goes through the same function. And `host-local`'s `cmdDel` releases by container ID plus interface name, so it needs nothing from the namespace. The new wording says only that a reboot may not deliver a `DEL` and that addresses can then stay allocated and shrink the usable range.

**On the node-local scope.** Also correct, and it is a gap that predates this change. The plugin's documentation puts it plainly: it "stores the state locally on the host filesystem, therefore ensuring uniqueness of IP addresses on a single host". The example hands one `rangeStart`/`rangeEnd` pair to every node that uses the `NetworkAttachmentDefinition`, and `br0` puts those nodes on one L2 segment. de76d37 adds a short paragraph after the `dataDir` one: give each node a non-overlapping range, or use a cluster-wide IPAM plugin such as `whereabouts`.

I kept this as a second commit rather than amending, so the change from the reviewed version is easy to read.

sterlingkoch (approved):

PR #762, opened 2026-09-14
docs: add note about omni users getting tel automatically

Iheanacho-ai:
## What

Add a how-to page covering Talos Enterprise Linux with Omni, and extend the Getting Started guide with a security scan step for enterprise accounts.

## Why

Closes #654. Omni accounts with a Talos Enterprise Linux entitlement get Enterprise Image Factory configured automatically — this was undocumented. The `omnictl security` commands for accessing vulnerability reports, SBOMs, and VEX documents were also missing from user-facing docs.

## Change

- Added `public/talos-enterprise-linux/how-to-guides/use-talos-enterprise-linux-with-omni.mdx` — covers how machines become Talos Enterprise Linux (schematic change trigger), installation media boot behavior, and a pointer to the Getting Started guide
- Updated `public/omni/getting-started/getting-started.mdx` — added Step 7 "Perform a security scan" with `omnictl security scan/sbom/vex --cluster-id` for Talos Enterprise Linux accounts; renumbered "Deploy your first workload" to Step 8

## Testing

- Ran: `omnictl security scan/sbom/vex --cluster-id` against a staging instance
- Where: `demo.omni.stage-pnap.managed.siderolabs.io`
- By: hand
- Result: staging instance does not have EIF configured — commands failed with stream closed; command syntax verified against `omnictl-1.12.0 --help` output; end-to-end test pending on an EIF-configured instance

PR #759, opened 2026-09-11
docs: add Getting Started with Talos Enterprise Linux guide

Iheanacho-ai:
## What

First-run tutorial for Talos Enterprise Linux, closing issue #657.

## Why

closes #657 

## Change

- Adds `tel/how-to-guides/getting-started-with-tel.mdx` with two paths: **Standalone Talos Linux** (UI and CLI tabs) and **Via Omni** (automatic delivery)

## Testing

- Ran: Steps 2–6 of the CLI tab (schematic creation, ISO download, SBOM/VEX download, grype scan, cosign verify-attestation)
- Where: staging (`factory-enterprise.staging.talos.dev`)
- By: author, by hand
- Result: all steps completed successfully; cosign staging identity differs from production by design
- Filled from outside the page: registry credentials (`--registry-username`/`--registry-password`) are required for cosign against the authenticated factory — discovered during testing and added to the guide

PR #765, opened 2026-09-14
docs: update omni CLI reference to v1.12.0 and replace deprecated omn…

Iheanacho-ai:
## What

Update the Omni CLI reference to v1.12.0 and replace all uses of the deprecated `omnictl download` command with the new `omnictl media preset create` + `omnictl media download` flow.

## Why

`omnictl download` is deprecated as of v1.12.0. The new `omnictl media` command group replaces it with a two-step preset-based flow. v1.12.0 also adds `omnictl security sbom/scan/vex` commands that were missing from the reference.

## Change

- Bumped `tools/omni-cli-gen/Dockerfile` from v1.9.0 to v1.12.0
- Regenerated `public/omni/reference/cli.mdx` — now includes `omnictl security` and `omnictl media` command groups; also regenerated `image-factory-configuration.mdx` and `omni-configuration.mdx` with updated schema
- Fixed `tools/mdx-normalize/main.go` to fence 4-space indented `Examples:` blocks in generated CLI docs, preventing em-dash rendering in Mintlify
- Replaced `omnictl download` with the new flow in: `getting-started.mdx`, `join-machines-to-omni.mdx`, `register-a-bare-metal-machine-iso.mdx`, `modify-kernel-arguments.mdx`, `override-ntp-servers.mdx`, `how-to-set-initial-machine-labels.mdx`, `how-to-register-an-aws-ec2-instance.mdx`

## Testing

- Ran: `omnictl media preset create`, `omnictl media download`, `omnictl media preset delete` with `--extra-kernel-args`, `--initial-labels`, `--platform aws`, and `--format pxe` variants
- Where: live Omni instance (`amara.na-west-1.omni.siderolabs.io`) using omnictl v1.12.0 binary
- By: by hand
- Result: all command variants succeeded; different schematic hashes confirmed flags were applied correctly


sterlingkoch (commented):
omnictl download is deprecated as of v1.12.0, and this PR removes it from every Omni page. Three of the TEL launch pages still reference it though, so they'll ship contradicting this one:

- manage-credentials.mdx line 14 (#753)
- migrate-clusters.mdx line 18 (#757)
- getting-started-with-talos-enterprise-linux.mdx line 173 (#759)

All three are prose references rather than command blocks, so it's a phrase swap to omnictl media download in each.

public/omni/omni-cluster-setup/how-to-set-initial-machine-labels.mdx
```diff
@@ -80,10 +73,11 @@ Initial labels work across most installation methods, including ISOs and PXE boo
 **For example:**
 
 ```bash
-omnictl download iso --initial-labels environment=production,region=us-west --pxe
+omnictl media preset create my-preset --arch amd64 --initial-labels environment=production,region=us-west
```
sterlingkoch: This creates my-preset again, and line 45 on this same page already created it. preset create goes through State().Create(), which returns AlreadyExists, so a reader working down the page gets a hard failure here. Same collision happens across pages: getting-started line 93, join-machines line 48, register-a-bare-metal-machine-iso line 23, modify-kernel-arguments line 83, and override-ntp-servers line 45 all create my-preset with different configurations.

For this page specifically, the PXE example only differs by output format, so it can reuse the preset from line 45 and drop the second create.

public/omni/getting-started/getting-started.mdx
```diff
@@ -87,17 +87,14 @@ You can download the Talos Omni image using either the CLI or the Omni UI:
 <Tabs>
   <Tab title="CLI">
 
-  Run the following command to download a Talos Omni ISO image:
+  First, create an installation media preset, then download from it:
 
   ```bash
-  omnictl download iso
+  omnictl media preset create my-preset --arch amd64
```
sterlingkoch: Two things, since this is the first place a reader meets presets.

omnictl media needs Omni v1.12.0, and the Omni docs are unversioned, so a self-hosted reader on v1.11 follows this and hits a command that doesn't exist. You gated omnictl security behind exactly this note in #762; media is on the primary getting-started path and wants the same one.

--arch already defaults to amd64 per the reference, so it's optional here and in the other five examples. Dropping it keeps the first command a reader ever runs shorter.

public/omni/omni-cluster-setup/registering-machines/how-to-register-an-aws-ec2-instance.mdx
```diff
@@ -185,16 +185,12 @@ You can download the Talos AWS image from the CLI or the UI:
...
 
 ```bash
-omnictl download aws
-```
-
-To see available options for adding system extensions:
-
-```bash
-omnictl download aws --help
+export ARCH=amd64  # or arm64 for Graviton instances
```
sterlingkoch: export ARCH=amd64 leaks a variable into the reader's shell for a value used once. --arch amd64 inline, with the Graviton caveat in the sentence above, reads simpler and matches the other pages.

sterlingkoch (commented):
A preset is a persistent server-side object, and these pages introduce it without saying so. Six pages create one, none mention that it sticks around, and omnictl media preset list / delete never appear in prose despite being in the reference. Two consequences worth deciding on together:

Names need to be unique per page, or reused deliberately. aws-preset on the AWS page is already the right instinct.
Pages that create a throwaway preset should say how to clean it up.
Where the same media is downloaded with different customizations, the alternative is one preset plus override flags at download time. omnictl media download accepts --extra-kernel-args, --initial-labels, --extensions and --arch directly, which would collapse the two-command block back to something closer to the old one-liner.

sterlingkoch (approved):

Iheanacho-ai:
superseded by #771

PR #774, opened 2026-09-17
docs: clarify Talos Enterprise Linux migration triggers and image source

### LINKED ISSUE #768: TEL migration guide: state that the move to Talos Enterprise Linux is one-way
Fast follow to [docs#757](https://github.com/siderolabs/docs/pull/757), which is being merged as part of the Talos Enterprise Linux launch section. Part of #649.

The migration how-to (`tel/how-to-guides/migrate-clusters-to-tel`) tells a customer how to move a cluster onto Talos Enterprise Linux, but does not tell them the move cannot be undone. That is the one thing a reader most needs before they trigger the upgrade, so it should be stated on the page rather than learned afterwards.

## What to add

A short note in the migration guide saying the move is one-way, placed before the steps rather than after them.

The mechanics, from [omni#2760](https://github.com/siderolabs/omni/issues/2760):

- Talos Enterprise Linux exists only for Talos v1.13.0 and later.
- Omni is connected to both factories and prefers the enterprise one for any version the two have in common. A user cannot choose the open-source factory for a version that Enterprise Image Factory also serves.
- Downgrading from Talos Enterprise Linux back to open-source Talos is prohibited.

So for a cluster already on v1.13.0 or later, the next upgrade of any kind moves it to enterprise images and there is no supported route back.

## Also worth considering for the same page

The same issue records a second consequence that is arguably sharper for the reader: if Omni loses its connection to the factory a machine's images came from, it treats those machines as orphaned and will not upgrade them. The only permitted operation is to destroy them. Whether that belongs in the migration guide or somewhere in the enterprise factory concept pages is a judgement call, but it is currently written down nowhere a customer will see it.

## Why it was not in the original PR

The point surfaced during launch-day discussion in `#proj-omni` on September 15, after the PR had been reviewed and approved. Splitting it out rather than holding the merge.

### LINKED ISSUE #769: TEL migration guide: recommend a throwaway kernel argument for clusters already on the newest version
Fast follow to [docs#757](https://github.com/siderolabs/docs/pull/757), which is being merged as part of the Talos Enterprise Linux launch section. Part of #649.

The migration how-to already says that any schematic change moves a cluster to enterprise images on the next upgrade, and it lists the options: bumping the Talos version, adding or removing a system extension, or changing kernel arguments. That is correct and does not need fixing.

What it does not do is tell the reader which of those to pick. Someone already running the newest Talos version has no version to bump to, and the obvious remaining move (add an extension, upgrade, remove the extension, upgrade again) is two upgrades and leaves them wondering whether the extension was supposed to stay.

## What to add

A sentence recommending a throwaway kernel argument as the cleanest trigger for a cluster already on the current version, and saying it can be left in place afterwards rather than needing a second upgrade to remove.

## Where this came from

Raised in `#proj-omni` on September 15. The first answer given was to add and then remove the `hello-world` extension; the better answer, from Utku, was to add a dummy kernel argument instead, precisely because it does not have to be taken back out.

This is the question the first internal person hit on launch day within minutes of the cutover, which is a reasonable signal that customers will hit it too.

Iheanacho-ai:
## What

Two additions to the Talos Enterprise Linux migration guide: a recommended trigger for clusters already on the newest Talos version, and a statement of which Image Factory Omni uses. Plus copy fixes in the same file.

## Why

Closes #769, closes #768. Fast follow to #757.

## Change

- Tip recommending a throwaway kernel argument, which can be left in place rather than removed by a second upgrade.
- Statement of which Image Factory Omni uses, scoped to Omni. The constraints in siderolabs/omni#2760 are Omni behaviours — it prefers Enterprise Image Factory for any version both serve, and offers no route back for a cluster — and don't apply to the `talosctl` path, where the installer image is named explicitly.
- Restructured the Omni section: the 1.13 threshold now appears before it's relied on, and three one-line paragraphs on image sources are merged into one.
- Two bare "factory" references replaced with full product names.
- Fixed a dangling colon in Step 2, where a caveat had been placed between the lead-in and its code block.

## Testing

- **Ran:** the four CI checks — `make broken-links`, `make validate-docs-nav`, `make style-check-changed STYLE_CHECK_BASE=upstream/main`, and `make docs.json && git diff --exit-code`. Separately, verified the recommended kernel argument on a live Omni instance.
- **Where:** local checkout for the CI checks; `amara.na-west-1.omni.siderolabs.io` (entitled) with two QEMU metal machines on Talos v1.14.0 for the argument.
- **By:** Claude Code, reviewed and edited by me; the Omni test run by hand.
- **Result:** all four checks pass, 0 style errors. One pre-existing sentence-case warning on line 55 — "Enterprise" isn't in `tools/style-guide-checker/exceptions.txt`, so it fires across several Talos Enterprise Linux pages; not addressed here. For the kernel argument: Image Factory accepted `talos.environment=ENTERPRISE_MIGRATION=true` at schematic creation, both machines booted from the resulting media and registered as connected, and `omnictl get kernelargs` shows the argument verbatim in `spec.args`. Machines from that media also came up labelled `enterprise`, confirming the page's existing claim about installation media from a provisioned account.
- **Not tested:** that adding the argument to an already-running machine triggers the upgrade which moves an open-source cluster to enterprise images. That scenario can't be reproduced on an entitled instance — Omni won't serve open-source images for 1.13+, and any upgrade that reaches the newest version already pulls enterprise, so the cluster arrives converted. It requires a cluster that predates the entitlement. The mechanism itself is the page's pre-existing text, attested in siderolabs/omni#2760.
- **Filled from outside the page:** the recommendation and its rationale come from `#proj-omni` (15 Sept); the Omni image-source mechanics from siderolabs/omni#2760, reviewed by @Unix4ever, who advised against describing what happens when an account loses its entitlement, as the behaviour is not currently defined; `talos.environment` confirmed against the Talos v1.14 kernel reference and `GetCmdline` in `internal/pkg/environment/environment.go`, which passes the value through unvalidated.


public/talos-enterprise-linux/working-with-talos-enterprise-linux/migrate-clusters.mdx
```diff
@@ -5,19 +5,21 @@ description: How to move existing Talos clusters and machines to Talos Enterpris
...
+Once your account is provisioned on Enterprise Image Factory, Omni switches each cluster to enterprise images on its next upgrade. No manual action is required beyond triggering that upgrade: any schematic change will do, whether bumping the Talos version, adding or removing a system extension, or changing kernel arguments. Enterprise images exist for Talos 1.13 or later, so a cluster on an earlier version, or one that has not yet been upgraded, continues running open-source Talos Linux.
 
-Enterprise images are available for Talos 1.13 and later. Clusters running earlier versions remain on open-source Talos Linux images until they reach that threshold.
+<Tip>
+If a cluster already runs the newest Talos Linux version, adding a kernel argument is the simplest way to move its machines to Talos Enterprise Linux: it changes the schematic, and it can be left in place afterwards. For example, `talos.environment=ENTERPRISE_MIGRATION=true`. See [Modify Kernel Arguments](../../omni/infrastructure-and-extensions/modify-kernel-arguments).
```
sterlingkoch, line 17: talos.environment=ENTERPRISE_MIGRATION=true works and does change the schematic, but it also sets a real process environment variable and the tip says to leave it in place afterwards. Are we comfortable publishing a permanent no-op variable as the recommended route, or should the version bump be the default and this the fallback?

Iheanacho-ai: so currently the version bump or any change to the schematic is the recommended path, i can make the tip a little more neutral-sounding and remove the example, but then it becomes a repetition of the paragraph before it.

i personally don't think the tip is very neccessary, and I think that Justin ran into this issue, because he had not seen the documentation at the time he was testing.

but let me know what you think @sterlingkoch

sterlingkoch: Agree, let's drop the tip. Line 14 already covers it. 

If you want to keep the newest-version case, add a clause there instead ("for a cluster already on the newest version, changing a kernel argument is the simplest trigger") without the example variable.

Issue #752, opened 2026-09-10
Broken vmware documentation for v1.14


gknepper:
here: https://docs.siderolabs.com/talos/v1.14/platform-specific-installations/virtualized-platforms/vmware

The first link is missing the content:

```
curl -fsSLO https://raw.githubusercontent.com/siderolabs/docs/main/static/talos/v1.14/vmware/cp.patch.yaml

```

here: https://github.com/siderolabs/docs/tree/main/static/talos

then using the cp.patch from v.13 `curl -fsSLO https://raw.githubusercontent.com/siderolabs/docs/main/static/talos/v1.13/vmware/cp.patch.yaml
`
Setting the cp.patch.yaml as:
```
- op: add
  path: /machine/network
  value:
    interfaces:
    - interface: eth0
      dhcp: true
      vip:
        ip: 192.168.1.222
```

and running `talosctl gen config vmware-test https://192.168.1.222:6443 --config-patch-control-plane @cp.patch.yaml` as explained in the documentation I got this error:

```
generating PKI and tokens
failed to generate config bundle: error patching control plane configs: JSON6902 patches are not supported for multi-document machine configuration
```

PR #776, opened 2026-09-17
fix: repair the broken v1.14 VMware and GCP install downloads

Iheanacho-ai:
## What

Restores the `static/talos/v1.14/` directory,.

## Why
Reported in #752.

## Change

- Add `static/talos/v1.14/vmware/`: `cp.patch.yaml` rewritten as a `Layer2VIPConfig` document, matching the [Virtual IP guide](https://docs.siderolabs.com/talos/v1.14/networking/advanced/vip); `vmware.sh` unchanged from v1.13.
- Add `static/talos/v1.14/gcp/`: `talos-ha.jinja` and `gcp-ccm.yaml` unchanged from v1.13; `config.yaml` with `talosVersion: v1.14.0`.
- VMware page: replace the JSON6902 example and wording with the strategic merge patch equivalent.
- VMware page: fix the scripted install, which downloaded `cp.patch.yaml` and saved it as `vmware.sh`.
- VMware page: drop shell prompts from command blocks so they can be copy-pasted.

## Testing

- **Ran:** `talosctl gen config` with the new patch, then `talosctl validate --mode cloud` on both generated files. Covers the VMware page's "Generating base configurations" and "Validate the configuration files" steps.
- **Where:** `ghcr.io/siderolabs/talosctl:v1.14.1` container.
- **Result:** Generation succeeds, and the `Layer2VIPConfig` document appears in `controlplane.yaml` with the expected VIP and link. Both validate commands report the file is valid for cloud mode. Confirmed `v1.14.0` is a published release and its Image Factory GCP image resolves.
- **Filled from outside the page:** dropped `dhcp: true` from the old patch after confirming the generated 1.14 config contains no link or DHCP documents, so DHCP is the default. Kept `eth0` after confirming cloud-target images get `net.ifnames=0`. The GCP manifests were not deployed — that change only restores missing files.


public/talos/v1.14/platform-specific-installations/virtualized-platforms/vmware.mdx
```diff
@@ -127,7 +122,7 @@ You can do this by issuing:
 
 <CodeBlock lang="sh">
   {`
-curl -fsSL "https://raw.githubusercontent.com/siderolabs/docs/main/static/talos/${version_v1_14}/vmware/cp.patch.yaml" | sed s/latest/${release_v1_14}/ > vmware.sh
+curl -fsSL "https://raw.githubusercontent.com/siderolabs/docs/main/static/talos/${version_v1_14}/vmware/vmware.sh" | sed s/latest/${release_v1_14}/ > vmware.sh
```
sterlingkoch, line 125: The redirect writes vmware.sh at 0644 and line 140 runs ./vmware.sh, so it needs a chmod +x. Same gap in 1.13, so fixing it here is fine.

sterlingkoch (commented):
The two red items from the accuracy bot are stale. It evaluated main, where static/talos/v1.14/ doesn't exist yet. Both factory URLs in the new assets return 200. Ignore them.

sterlingkoch (approved):

PR #777, opened 2026-09-18
docs: remove v1.6 and deprecate v1.7 talos docs

Iheanacho-ai:
## What

deprecated the 1.7 version and deleted the 1.6 version of the talos docs.

## Why

in keeping with a cadence established about cycling through the versions

## Testing

ran `make preview`


sterlingkoch (commented):
Mechanism looks right and matches how v1.6 was deprecated: out of docs.json and the Makefile lists, files kept, pages still serving. Two things. The v1.6 URLs are live today and will 404 after this, so do we want /talos/v1.6/* redirected to v1.14, or is a 404 fine for a version this far past EOL? 
And common.yaml:92-93 still carries a v1.6 redirect whose destination is about to disappear, worth deleting here.

sterlingkoch (commented):
Redirects look good. The wildcard covers the 404 question, and the two that land in 1.7 are right since that's the newest version with those pages.

This conflicts with main now (it predates #783), so it needs a rebase and a regenerated docs.json.
When 1.6 was hidden, talos-v1.6.yaml stayed in the repo and only came out of the Makefile. Can we do the same with talos-v1.7.yaml instead of deleting it? Makes it easy to restore.
#788 edits a 1.6 file that this deletes, so whichever merges second will conflict. I'd merge #788 first and let the delete win on rebase.
Could the description say which cadence this is? A 50k-line deletion is worth one sentence for whoever reads the history later.

Issue #724, opened 2026-09-01
Retitle and re-scope the system extension authoring guide


sterlingkoch:
## Problem

The complete guide to authoring a Talos system extension is published as [Adding a Kernel Module](https://docs.siderolabs.com/talos/latest/build-and-extend-talos/custom-images-and-development/kernel-module). Despite the title, that page covers the entire authoring path: creating a package in `pkgs`, the `bldr` `pkg.yaml`, creating the extension, the `manifest.yaml`, building the installer with `imager`, and installing on a fresh or existing machine. A kernel module is one worked example of that path, not the scope of it.

Meanwhile the "Author system extensions" section of the [System Extensions](https://docs.siderolabs.com/talos/latest/build-and-extend-talos/custom-images-and-development/system-extensions) page is three sentences:

> A Talos system extension is a container image with the specific folder structure. System extensions can be built and managed using any tool that produces container images, e.g. `docker build`. Sidero Labs maintains a repository of system extensions.

So a reader who wants to build an extension opens the page named for their task, gets three sentences and a pointer to GitHub, and never reaches the guide that would have answered them. This surfaced in user feedback in August: a user could not find extension authoring documentation and is running Talos in VMs as a result.

## What to do

1. Retitle the kernel-module page to name what it actually covers (something like "Building a system extension"), keeping the kernel module as the worked example. Add a `redirects` entry for the old path.
2. Cross-link it from the "Author system extensions" section of the System Extensions page, as the primary path rather than the GitHub pointer.
3. Bring in the two things that currently live only in the [extensions repo README](https://github.com/siderolabs/extensions#building-extensions): the `manifest.yaml` structure, and the `rootfs` path allowlist (extension files are confined to a specific set of paths, no special files, no world-writeable files). A reader who follows our guide and then hits an unstated path restriction has to go read the repo anyway, which defeats the point of having the guide.
[...]

PR #778, opened 2026-09-21
docs: rescope the system extension authoring guide

Iheanacho-ai:
## What

Renames "Adding a Kernel Module" to "Building a System Extension" and reframes it around the whole authoring path, keeping the kernel module as the worked example. Adds the extension structure, the `rootfs` restrictions, and the public Image Factory constraint. Links it from the System Extensions page.

## Why

The complete guide to authoring a system extension was published under a title naming only one use case for it. Readers looking for extension authoring landed on the System Extensions page instead, whose "Author system extensions" section was three sentences and a link to GitHub.

Reported in #724, prompted by user feedback from someone who could not find the documentation and ran Talos in VMs as a result.

## Change

- Rename `kernel-module.mdx` to `building-a-system-extension.mdx`, with a redirect from the old path.
- Reframe the intro around the full path; the kernel build is now presented as a kernel-module requirement rather than a universal first step.
- Add the extension structure and the `rootfs` path restrictions, previously only in the extensions repo README.
- Add the `extensions-validator` test step to the worked example, and document what it catches.
- Note that the public Image Factory builds only from the official extension catalog, so a custom extension needs `imager`.
- Link the guide from the System Extensions page.

The walkthrough itself is unchanged. Four lines were removed from the old intro; everything else is a rename or an addition.

## Testing

- **Ran:** built two extensions against `siderolabs/extensions` on `linux/arm64`, then `imager metal --arch arm64 --system-extension-image`. Covers the extension structure, the `rootfs` restrictions and the installer build.
- **Where:** Colima, standalone buildx builder, local registry.
- **Result:** an extension writing to `/usr/local/bin` passes the validator; one writing to `/opt/foo` fails with `path "/opt/foo/docs-test" is not allowed in extensions`. Without the validator step the same extension builds and `imager` packages it into an image without complaint, which is why the guide now includes the step. The public Image Factory returns HTTP 400 for a non-catalog extension: `official extension "..." is not available for Talos version v1.14.0`.
- **Filled from outside the page:** the allowed-path list comes from `AllowedPaths` in `pkg/machinery/extensions/extensions.go`, not the extensions README, which is abridged. `manifest.yaml` fields verified against the `Metadata` struct. The kernel build and module signing were not tested.


public/talos/v1.14/build-and-extend-talos/custom-images-and-development/building-a-system-extension.mdx
```diff
@@ -1,18 +1,82 @@
...
 
-[System extensions](./system-extensions) in Talos provide ways to add files to the root filesystem and make it possible to run privileged containers as services.
-But Talos still requires that kernel modules be signed with a trusted signing key to be loaded at run time.
+[System extensions](./system-extensions) add files to the Talos Linux root filesystem and make it possible to run privileged containers as services.
 
-To add a kernel module to Talos you will need to create a system extension that is built with the kernel and signed by the same signing key.
-System extensions without kernel modules work without this requirement, but there are a couple extra steps needed when adding kernel modules.
+This guide builds a system extension, packages it into an installer image, and installs it on a machine.
+The worked example is a kernel module, which additionally requires a custom kernel build.
+An extension that ships no kernel modules does not need one: start at [Create an extension](#create-an-extension).
```
sterlingkoch, line 15: Lines 15 and 79 send a non-module extension to "Create an extension," but that section builds against ${PKG_IMAGE} from the kernel step and copies /usr/lib/modules (line 235), so it can't be followed if you skipped ahead. Either cut those two sentences or add a plain binary extension before the kernel-module example. Rest of the rescope looks good, and the rootfs allowlist matches AllowedPaths in 1.14.1.

sterlingkoch (dismissed):
controlplane.yaml and worker.yaml at the repo root look like leftover gen config output from testing. They've got CA keys and tokens in them, and this repo is public, so they need to come out before merge (squash would keep them). Might be worth adding controlplane.yaml / worker.yaml / talosconfig to .gitignore too, since this'll keep happening with agent testing.

Content looks good. I checked the rules against WriteUserFiles in 1.13 and they match.

sterlingkoch (commented):
The rename leaves the old kernel-module pages in 1.7 through 1.13 with canonical links pointing at the v1.14 kernel-module URL, which now redirects. make canonical-links-check flags all seven. make canonical-links-local fixes them. CI doesn't run that check, which is why it's green.

Issue #721, opened 2026-08-27
Redistribute coreDNS pods after fresh cluster install


zenhighzer:
## Description

When bootstrapping an HA Talos cluster, both CoreDNS replicas may initially be scheduled on the same control plane node. If that node crashes, DNS resolution is not available until the pod(s) are rescheduled.

<img width="1796" height="558" alt="Image" src="https://github.com/user-attachments/assets/554c1a51-6c2b-4aa8-b8d5-37b551c7bc12" />

The Kubernetes kubeadm HA documentation recommends restarting CoreDNS after the additional nodes have joined:

* https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/#steps-for-the-rest-of-the-control-plane-nodes

```bash
kubectl rollout restart deployment -n kube-system coredns
```

This allows the scheduler to redistribute the CoreDNS replicas across the available nodes.

<img width="1789" height="524" alt="Image" src="https://github.com/user-attachments/assets/c3eb9a47-067e-4bcc-8442-a27b466ff82c" />

It might be useful to add a similar note to the Talos HA/production cluster documentation.

Related Kubernetes documentation issue: kubernetes/website#52089

## References

* https://github.com/kubernetes/website/issues/52089
* https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/#steps-for-the-rest-of-the-control-plane-nodes

Iheanacho-ai:
Hi @zenhighzer, thanks for the report!

We won't document the restart, though, because it doesn't stick. Talos only tells the scheduler to *prefer* different nodes, and that preference is applied once when a pod is placed. A restart shuffles the pods, but the next eviction or reboot can put them back together. In our testing, the same restart command spread them in one run and co-located them in another.

If you need them guaranteed apart, disable the built-in CoreDNS and deploy your own with affinity rules suited to your cluster:

```yaml
apiVersion: v1alpha1
kind: KubeCoreDNSConfig
enabled: false
```

PR #781, opened 2026-09-21
docs: add a step to spread the CoreDNS pods across nodes

Iheanacho-ai:
## What

Adds a step to the production cluster guide covering CoreDNS pod placement after bootstrap.

## Why

CoreDNS runs two replicas, both scheduled during the bootstrap while only the first control plane node is available. The deployment prefers to spread them across nodes but does not require it, and the placement is not revisited once the pods are running. Both replicas can stay on one node, making cluster DNS a single point of failure in a cluster built for HA.

Reported in #721.

## Change

One step added after "Verify your nodes are running": check where the CoreDNS pods are, and if both are on the same node, delete one so the scheduler places the replacement on the empty node.

## Testing

- **Ran:** the full production cluster guide, Steps 5 through 16, then the new step.
- **Where:** two Talos v1.14.1 control plane nodes in UTM.
- **Result:** both CoreDNS pods scheduled onto the first node and stayed there for six minutes after the second node became Ready, confirming the behaviour. Deleting one pod placed the replacement on the empty node, giving one ready DNS endpoint per node.
- **Filled from outside the page:** `kubectl rollout restart` was tried first and is not reliable — it replaces both pods, and mid-rollout each node holds one, leaving the scheduler no preference. It moved both replicas onto one node in one run and spread them in another. Deleting a single pod leaves an unambiguous preference, so the guide uses that. A hard `topologySpreadConstraints` was also tested: it guarantees spread but leaves a replica `Pending` when nodes are short, so it is not recommended here.


sterlingkoch (approved):

public/talos/v1.14/getting-started/prodnotes.mdx
```diff
@@ -394,6 +394,29 @@ Run the command to ensure that your nodes are running:
 kubectl get nodes
 ```
 
+## Step 17: Spread the CoreDNS pods across nodes
+
+CoreDNS runs two replicas by default.
```
smira, line 399: please don't put this, this is wrong on any level.

Iheanacho-ai: hey @smira sure, ill drop it.

can i ask what is wrong with this addition. I ask because I did reproduce the symptom on a two-node cluster: both CoreDNS replicas were scheduled onto the first control plane node during bootstrap and were still there even after the second node became Ready.

smira: yes, because you need to look into anti-affinity in Kubernetes. Talos default CoreDNS manifest is very conservative as it doesn't know how many nodes the cluster has.

If one wants to spread - they should deploy their own CoreDNS with proper affinity set up to schedule across multiple nodes.

what you proposed is wrong, as making the pods randomly land on different nodes doesn't mean it would persist - next scheduling decision might put them on the same node

(you should review with Claude, it can explain it better than I possibly can)

And putting this into "Production Notes" is:

a) wrong advice
b) showing that we don't know how pod scheduling works

Iheanacho-ai: Okay, that makes sense. I will close this PR and ask the user to deploy their own coredns. Claude explained and agreed with this premise, which is incredible, seeing as it recommended this solution in the PR in the first place as way to kick off talos soft anti-affinity

PR #782, opened 2026-09-22
Adopt the Talos design system tokens

sterlingkoch:
First of three docs-site changes bringing the site onto the shared design system ([siderolabs/talos-design-system](https://github.com/siderolabs/talos-design-system)).

**What changes**

- `public/talos-tokens.css`: the generated Mintlify build of the token set (colors, spacing, type, Mintlify theme variables, `@font-face` rules). Vendored, not imported; the header explains the update procedure.
- `public/fonts/`: self-hosted Manrope and JetBrains Mono (variable-weight woff2, latin + latin-ext) with their OFL licenses.
- `docs.json` colors: the pink-magenta trio becomes the brand red (`#C44740` primary, checked to AA against white labels; the marketing-site `#E04B45` itself does not pass under white text, so the token set carries a darkened fill variant).
- `style.css`: the hand-written table and layout rules stay, with their raw grey hexes re-pointed at tokens. The announcement banner picks up the brand gradient (the AA-safe fill variant).
- Banner copy: en dash to hyphen.

**What does not change**

No content, navigation, or page changes. The landing page and IA work come separately.

**Review**

The Mintlify preview deployment on this PR is the thing to look at: check the accent in both themes, the type change (Manrope/JetBrains Mono replacing Mintlify's defaults), tables, and the banner.


Iheanacho-ai:
review from claude:

Description says #C44740 but the diff sets #B93B46. Both pass AA (4.86:1 and 5.56:1 against white) and the #E04B45 point is right at 3.99:1, so this is just the description being stale.

Iheanacho-ai (approved):

PR #783, opened 2026-09-22
Landing page: generated homepage and products navigation

sterlingkoch:
Replaces the docs site root with a landing page in the design system, and switches the navigation from tabs to Mintlify **products**.

## What this does

- **Navigation: tabs → products.** Every product gets its own sidebar and version switcher, and the navbar gains a product switcher. The page set in `docs.json` is byte-identical to before; only the structure changed.
- **The homepage is generated.** `public/index.mdx` is written by `tools/docs-gen` from a `homepage:` block in `common.yaml` plus the `family` each product declares in its nav file. Adding a product adds its card; a versioned product's card meta line ("v1.14 current") tracks its newest version automatically. `make docs.json` regenerates it and CI fails if the committed copy is stale.
- **Announcement gating is data.** Announced products without documentation (Talos Linux Hypervisor, Talos Linux Containers) are declared as `soon: true` cards: present, "Coming soon", not clickable. Unannounced names sit as commented-out YAML in `common.yaml` with a note on when to flip them.
- **Contextual navbar chrome** (`public/script/contextual-chrome.js`): the search placeholder names the product being read, the product switcher is labelled "Product", and the version switcher is hidden on the landing page. The call to action stays global (`navbar.primary`).
- **Changelog** becomes a hidden product with its sidebar suppressed via generated `product-chrome.css`, reached from the homepage's "Also" row as Release notes.
- **CI** builds the generator from the in-tree source (`make docs.json-local`) so a PR that changes the generator is checked against its own version, and `tools/docs-gen/**` joins the path filters.

## Design

The page itself is the approved prototype: hero with a search affordance that opens the Mintlify search modal (`public/home.js`), domain sections (Kubernetes live, Hypervisor and Containers coming soon), a lighter Tools strip, and an "Also" row. All styles are scoped under `#docs-home` and use only design-system tokens (`public/home.css`).

The generator architecture originates in `siderolabs/docs-staging` (Amarachi's); this PR ports it and points the template at the approved design, so the two repos can converge instead of carrying two homepages.

Built on #782 (design-system tokens); the diff shrinks once that merges.


sterlingkoch:
Retargeted to main so Mintlify builds a preview (it skips PRs against non-deployment branches). Until #782 merges, this diff includes its commits; once it merges, this collapses to just the homepage.

Iheanacho-ai:
Two things from me:

- I see this drops the per-product CTA logic from the staging version. Is that deliberate? The one thing it buys is that the global button reads "Try Talos Omni" on Talos Enterprise Linux and Kubernetes Guides pages, where it's a different product.

- I also updated the contributing guides in the staging repo, since these changes make a fair bit of the contributing docs stale. Happy to port those across once this lands.

A few things Claude picked up reviewing the diff:

- `soon` card with no `tag` never shows "Coming soon". The pill renders inside the `card.Tag != ""` block, so a card without a tag gets the dashed border but no label. Every card here sets one, so it's fine today. It would bite the first time we add a `soon:` product whose licensing isn't decided yet.
- Card fields are interpolated straight into MDX. A `{` in a title or description would start a JSX expression and break the page build. Unlikely with prose, but worth knowing the failure is a broken build rather than a mangled character.
- Nit: `home.js` doesn't guard `document.body` before calling `observe()`, where `contextual-chrome.js` guards with `readyState`. Not reachable today, since Mintlify injects scripts in the body. Just inconsistent between two files added together.

Iheanacho-ai (approved):

